PR #95, opened 2015-01-25
Experimental plex support (still no configurable)

Hairo:
Added experimental [plex media server](https://plex.tv/) support (need to manually set the configs for now, check the changes to utils.py) based off [one of my projects](https://github.com/Hairo/saya).

### DIFF
--- a/trackma/lib/libplex.py
+++ b/trackma/lib/libplex.py
@@ -1,0 +1,67 @@
+#!/usr/bin/python
+
+import trackma.utils as utils
+import re, ntpath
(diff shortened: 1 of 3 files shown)


z411:
Thank you very much. I can't test this, but since Plex tracking seems to be an alternative to local tracking, I think it would be better to have a `tracker_type = 'local' or 'plex'` config option (with local as default) instead of `plex_enabled`. Is it okay for me to make that change?

Hairo:
No problem, something like that?

z411:
Great, thank you.

PR #109, opened 2015-03-23
Make play_episode work with multi-episode files

awlayton:
Hopefully these changes aren't horrendous or anything.

1. Changes `_search_video` to look for a file ***containing*** `episode` rather than only one ***ending*** with `episode`
1. Changes `play_episode` to return the last episode of the played file

PR #111, opened 2015-04-02
Fix check for if an episode is in the range of episodes in a file.

awlayton:
This should fix the bug introduced by #109.


z411:
Thank you.

PR #112, opened 2015-04-19
Remember geometry of Qt GUI when closing the application

veluria:
Something I proposed in the past. Upon closing the application, the geometry is stored. I'm not sure whether I'm using the config file the right way, so feel free to correct (or reject) it if you don't like it.


z411:
Thank you, it looks fine. But shouldn't we let the window manager handle this?

veluria:
You are right. Some window managers have few to zero options regarding the positioning, though. I turned it into an option. How about this?

z411:
Thank you.

PR #114, opened 2015-04-27
Changed left/right description in help for curses client and improved colours

anjandev:
I had trouble finding out how to change my filter at the top. When I found out, I thought the help description of left/right controls were not very good.

I made the description for the left/right keys better in the help for the curses client.

Also, the yellow colours made it very diffucult to read the headers in the curses client. So I made the header text white.


z411:
Thanks.

PR #125, opened 2015-06-28
Removed trailing whitespace

rr-:
To make sure no other changes happened, I recommend viewing the diff with `?w` option:

- https://github.com/z411/trackma/pull/125/files - normal diff
- https://github.com/z411/trackma/pull/125/files?w=1 - diff without whitespace changes

### DIFF
--- a/trackma/accounts.py
+++ b/trackma/accounts.py
@@ -4,10 +4,10 @@ import cPickle
 class AccountManager(object):
     """
     This is the account manager.
-    
+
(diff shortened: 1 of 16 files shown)


JeroenMathon:
What i'm about to say is off topic but,
One of the many reasons why i choose C and C++ over Python is the white spaces(Python really pisses me off sometimes)
Other than that good job on cleaning up the code ^_^

z411:
Thank you.

PR #127, opened 2015-07-07
Fix a bug where a single result wouldn't be selected in search

veluria:
This simply clears the current selection before a search occurs. If only a single result was present, it became impossible to select it as it seemed to be selected already. Sorry for the merge commits, forgot to rebase. ^^;

Edit: It works if you simply click another column. Kinda confusing, though.


z411:
Thanks

PR #130, opened 2015-07-29
Anilist status fixes

mutilx9:
This fixes two bugs at libanilist.

Parsing bug: in the parsing function, instead of using the string that change with the mediatype (self.airing_str), 'airing_status' is used. This prevents the manga statuses to be parsed.

Retrieve bug: anilist added some months ago the cancelled status, if you have in your list one item that has this status, trackma raises a KeyError. I set cancelled to 4, but 2 could do, cancelled is like finished.


z411:
Thank you!

PR #134, opened 2015-08-10
CLI: Fixed typo that breaks sort command

mutilx9:
The cli's sort command doesn't work because of a typo, this commit fixes it.


z411:
Thank you.

PR #147, opened 2015-09-11
Added more key bindings for gtkui

hbers:
After switching from curses ui to gtk I noticed there aren't many key bindings for the actions. So I added more.


z411:
This makes it more usable, thank you.

PR #155, opened 2015-10-15
[Qt] Minor enhancements

Birdulon:
Reduces some deadspace, enables the wonderful percent widget on shows that don't have a confirmed number of max episodes.


z411:
Thank you!

PR #156, opened 2015-10-18
[Qt] Customisable row highlights

Birdulon:
Allows user customisation of row highlight colors. Supports arbitrary colors as well as colors from the Qt theme (options dialog only shows background roles since foreground roles are very likely to have little contrast to the text, but any valid roles can be entered in the config json).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,6 +18,7 @@ rhumroux
 Zdmshadow
 loimu
 rr-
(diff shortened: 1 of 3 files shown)


z411:
Great work, thanks.

Birdulon:
(づ｡◕‿‿◕｡)づ

PR #159, opened 2015-10-25
[Qt][Option] Use the current API icon as the tray icon

Birdulon:
Should make running multiple instances for different APIs a bit more manageable.


z411:
Oh nice, thanks. Icons are currently pretty small so I might get higher resolution ones if available.

PR #161, opened 2015-10-30
libanilist: Fixed an exception when trying to delete

mutilx9:
The anilist API returns an empty response when you delete an entry. At the _request function, the response is parsed and raise a ValueError. The entry is deleted at anilist, but trackma crashes.

I used a try/except just like in the search function. First, I did `if method == "DELETE"` in _request, but I think it's better this way because it only runs when the delete function is called.


z411:
Thank you.

PR #163, opened 2015-11-02
[Anilist] Additional info fields; Start date on lists

Birdulon:
Reordered the info and renamed some data to be more in line with libmal. Uses the merge method libvndb does to get start date on the list views (yay, estimated eps) so first run will take a while on accounts with a fair bit of anime. Old anime.info files will prevent loading the new fields, considering adding a button in Qt options to purge old db but I'm open to other suggestions in dealing with stale dbs.

### DIFF
--- a/trackma/lib/libanilist.py
+++ b/trackma/lib/libanilist.py
@@ -247,6 +247,7 @@ class libanilist(lib):
                     'total': self._c(item[self.mediatype][self.total_str]),
                     'image': item[self.mediatype]['image_url_lge'],
                     'image_thumb': item[self.mediatype]['image_url_med'],
+                    'url': str("http://anilist.co/%s/%d" % (self.mediatype, showid)),
                 })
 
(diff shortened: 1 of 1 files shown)


z411:
Thank you. Sadly I'm not sure if I can merge this.

The merge functionality works (and is necessary) on VNDB because you can pull the detailed info of all VNs in your list at once with a few paged requests, so it's fast and requires one request per 25 VNs.
In this case however, it's one extra request per one item, which I think is too much. My list has 500+ shows and not only it takes *too* long the first retrieve, but puts a considerable strain on the server too. We're thankful this kind of thing doesn't get us banned yet!

Birdulon:
Fair enough, I'll see if I can make use of the airing data from browse requests.

Any comments on #164?

Birdulon:
Closing this for now, I'll open a new PR when I get the episode data represented without overburdening the API. :)

Issue #153, opened 2015-10-12
Progress bars appearance [GTK]


archnous:
The progress bars in 0.3 were better looking, especially in dark themes. I haven't checked the code, but I believe in 0.3 they followed the theme's widget style, instead of being arbitrarily set as in 0.4 (please, correct me if I'm wrong).

![trackma-0 3](https://cloud.githubusercontent.com/assets/6310912/10425203/4f68bda4-70de-11e5-8c6a-1844bc8a5fa0.png)
![trackma-0 4](https://cloud.githubusercontent.com/assets/6310912/10425204/4f6920d2-70de-11e5-8084-d5a967c5a313.png)

wizbright:
![151012112642-572x545](https://cloud.githubusercontent.com/assets/1281756/10432965/3da0f2f8-70d4-11e5-8d6e-5ddd97af4443.png)

I guess it depends on the theme. I think the bars look fine to be honest.

archnous:
I'd prefer the uniform look of the theme. Right now the colour of the bars is fixed; please see how much more elegant is the look of the themes, whether they're dark or light:

![2015-10-12-230707_1920x1080_scrot](https://cloud.githubusercontent.com/assets/6310912/10437675/69d99a92-7137-11e5-9e75-24b6348705cb.png)
![2015-10-12-230814_1920x1080_scrot](https://cloud.githubusercontent.com/assets/6310912/10437676/69dc7b90-7137-11e5-9256-c02033275286.png)
![2015-10-12-230901_1920x1080_scrot](https://cloud.githubusercontent.com/assets/6310912/10437672/69d0d790-7137-11e5-95d2-cf492d981c18.png)
![2015-10-12-230930_1920x1080_scrot](https://cloud.githubusercontent.com/assets/6310912/10437674/69d65468-7137-11e5-8316-d29f27361039.png)
![2015-10-12-230958_1920x1080_scrot](https://cloud.githubusercontent.com/assets/6310912/10437673/69d144b4-7137-11e5-9fd7-0e7553134c0b.png)

I believe this should be reverted to the old behaviour, i.e. let the theme apply its colours.

z411:
The new progress bar is made from scratch since it also shows aired episodes and available episodes in your hard disk (try List -> Re-scan library). Reverting the changes would also destroy this functionality.
The best solution is to modify the current custom progress bar to use the theme colors. I'll be confirming this.

PR #164, opened 2015-11-02
[Qt,Utils] Play consolidation; purge account db button

Birdulon:
Combined "Play" and "Play Next" buttons into a button menu which defaults to Play Next and additionally has "Play Episode..." for playing arbitrary episode numbers. Added sorting on the percent column of the shows table, and a context menu containing a subset of the Show menu.
Added a Purge button to the account manager for discarding stale dbs (mostly for #163).


Birdulon:
Added progress/percent widget color selection, fixed colors updating on settings dialog cancellation, added a nested dict update in parse_config to maintain compatibility with older configs that lack the additional color fields.

Birdulon:
(Inspired by #153) Added an EpisodeBar style which overlays the episode sub bar onto a standard-looking ProgressBar to keep closer to the Qt widget style look. Added relevant settings to allow choice of the standard ProgressBar, the current dual bar or this new dual bar. Settings dialog started to get too tall so I moved the EpisodeBar and Color settings to a new tab.

z411:
Thank you for your work. I'll be merging most of this, but before merging I'd like to comment a few points:

1. Instead of having a numeric entry dialog to specify the episode to play, what about generating he popup menu according to the number of episodes? The GTK does this when you right-click a show in the list.
2. I don't use KDE and don't get any icons, so we should probably provide stock icons. I've wanted this for some time anyway because I wanted to add a toolbar, and icons always make a GUI look better.
3. About the purge button; purging the database can already be done automatically by bumping the version variable in the data.py file (line 40). Should we still provide manual purging?

I can do most of these changes, don't worry.

Birdulon:
1) sounds good to me, I was thinking of doing that before I got bogged down
in what to do with shows that have hundreds of episodes (Gintama) and went
with numeric entry as an easy out.

2) yep yep

3) I didn't know that. From a developer standpoint a manual button is nice
but if it's useless to users it could be omitted or concealed.

Cheers

z411:
Okay thanks, merged.

PR #166, opened 2015-11-13
[Anilist][Utils][Qt] Anilist improvements

Birdulon:
Reordered the info and renamed some data to be more in line with libmal, fixed URL generation. Uses airing data from the anilist browse API (1 request per sync, ~150kB json) to determine number of episodes aired for airing shows. Replaces defunct (start-)date column for anilist anime with a simple days,hours countdown to the next episode (only updates on table regeneration, I may revisit this later).

Anilist API version bump from 1 to 1.1


z411:
Thank you.

Issue #140, opened 2015-09-02
Remember window size in trackma-gtk


dyskette:
Currently trackma-gtk does not remember the last size of the window used, so when is opened again it needs to be resized. The interface should remember this user preference.

dyskette:
Fixed. 

Thanks @Birdulon.

PR #167, opened 2015-11-13
[Gtk] Minor additional options

Birdulon:
Added options for remembering window geometry (#140), using standard progress bar widgets (#153) and using API icons in the tray (port of #159). Fixed percent sort.

I'm giving up on overlaying the episode subbar onto the standard widget in Gtk, but I'll probably get around to adding a theme-aware colour picker within the next week.


z411:
Thank you.

Issue #132, opened 2015-08-01
Right click > Play episode # / Status: rewatching


anjandev:
It would be nice to be able to right click a show and select an episode to watch like you can with taiga. Watching episodes from something that you have copmpleted is very cumbersome with the current ui. 

Similarily, it would be nice to add a rewatching status.

z411:
I see, I haven't checked but what happens when a show has 100+ episodes? Is it a scrolling list?

anjandev:
I think so, if u would move ur mouse to the lower part of the list, it would scroll.

z411:
This was implemented in the GTK interface. Please tell me if this is what you wanted so I can implement it in the Qt interface too. Any other suggestions to improve the (currently clunky) interfaces are also welcome.

anjandev:
well, taiga recently changed their select episode to play option to look like this:
![trackma](https://cloud.githubusercontent.com/assets/8602360/9692991/bd911696-5301-11e5-84b7-201bacf2f0a6.png)
I feel like this is a better ui choice than one giant scrolling list.

veluria:
@z411 
There is a logic error in your code.

    for i in xrange(1, show['total']):

Should be

    for i in xrange(1, show['total'] + 1):

instead.

z411:
You're right @veluria, thank you.

z411:
This was implemented in Qt by Birdulon.
For the rewatching status refer to #141

PR #168, opened 2015-11-14
[Qt] Play episode submenu selection

Birdulon:
Replaces episode selection dialog with a series of submenus. Longer series will have more levels of submenus. I've used breakpoints that I feel strike a good balance of menu size and levels of submenus, but if you wish to tweak them the variables are labelled and at the top of the generate_episode_menus method. Unreasonably long shows will fall back to the numeric dialog, but with the supplied breakpoints you'd be hard pressed to find a show with that many episodes :3c
Also changed the aired eps estimation function to disregard totals of 0.
Related Issues: #132, #164
trackmaplayep1
trackmaplayep2
trackmaplayep
trackmaplayep3


z411:
Thank you.

Issue #39, opened 2014-04-28
Add "airing date" columns to the table for MAL anime|manga lists


YogSottot:
Add "airing date" columns to the table for MAL anime|manga lists. With sorting function.
![airing_date](https://cloud.githubusercontent.com/assets/7411302/2816345/46bb56ee-cec8-11e3-893a-c604c61675e8.jpg)

z411:
Implemented.

PR #172, opened 2015-11-19
[GUIs][Utils] Date display rework

Birdulon:
Adds start date and next episode ETA columns for Gtk, Curses on supported APIs. As I only use MAL and Anilist, I took conservative guesses for the other APIs (i.e. if I didn't see start_date set, I assumed it wasn't there). A bit of refactoring with the existing Qt functionality. This should complete #39.


Birdulon:
Obviated by b6616c8 to 08e361b.

PR #183, opened 2015-12-09
[Qt] Fixed countdown cells having no background …

Birdulon:
…when countdown unavailable. Centered countdown and date column text.
Stops the items with no data from looking too out-of-place.


z411:
Thank you.

PR #184, opened 2015-12-21
[Engine][libanilist] Added shows default to PTW, anilist bugfix

Birdulon:
Having shows add straight to PTW is convenient as often shows are added to userlists ahead of time, and they move to Watching when the tracker picks them up.
Anilist omits placeholder data on the total field for some shows, messing up the UI when those shows are added. 1263516 should not only fix this issue but any other omissions.

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -342,8 +342,12 @@ class Engine:
         if status:
             if status not in self.mediainfo['statuses']:
                 raise utils.EngineError('Invalid status.')
+            else:
+                show['my_status'] = status
+        else:
+            # Default add to 'plan to read/watch'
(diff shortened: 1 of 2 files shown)


z411:
Hi, I merged the bugfix, thank you.
But I didn't merge the PTW default change, because Trackma should actually add the show to the list currently selected (for example if I open the On Hold list, and I open the Add dialog, it should add the show to the On Hold list).
This actually works in the curses UI, but it's not working in the Qt and GTK interfaces. I'll fix that. Thanks.

Birdulon:
No worries, cheers!

z411:
Done.

Issue #141, opened 2015-09-05
Show priority of watching show in Plan to Wartch


anjandev:
Mal has an option to select your priority for shows in your plan to watch list section. It would be nice if trackma could show me this in the gtk/qt guis.

z411:
I wasn't aware of this. Implementing this might be useful because VNDB also has a similar Wishlist with priorities. Do you have any suggestions on how it could be implemented in the user interfaces?

anjandev:
I would like to sort my shows in the plan to watch by priority. It would be nice to have a seperate column for priority that only appears when I click the Plan to Watch tab.

anjandev:
Similarily, it would nice to have more options that I could edit on mal from trackma. Suchas, this is mal when I click edit from their site:

![2015-09-04-182404_3600x1080_scrot](https://cloud.githubusercontent.com/assets/8602360/9697017/4190d776-5332-11e5-9ccd-bf439fa62eb2.png)

Things like start date and end date could be automated to be pushed. Suchas when I start an anime, trackma could automatically send mal my start date. Also, since trackma keeps track of how many episodes have been downloaded it could update these as well.

It also would be nice to have columns within the guis that can be enabled/disabled for these extra attributes.

z411:
Yeah, I think it's time to make all these options visible and modifiable. I'm just not sure of two things:

1. How to implement them in the user interfaces; one idea is to add them in the Show details window when you double click the show, like Taiga does. Another idea is to add a Show detailed options button in the main window.
2. More importantly, not all sites have the exact same options. We have to find a way to make it cross-compatible between different sites (AniDB, MAL, VNDB, Hummingbird).

>Things like start date and end date could be automated to be pushed. Suchas when I start an anime, trackma could automatically send mal my start date.
Trackma already does this internally, it's just not shown anywhere. The settings dialog of the Qt interface has the option "Change start and finish dates automatically" in the Sync tab, it should be enabled by default.

anjandev:
>1. How to implement them in the user interfaces; one idea is to add them in the Show details window when you double click the show, like Taiga does. Another idea is to add a Show detailed options expand arrow in the main window.

I think doing how taiga does it is prefect. Trackma could have two tabs when the new box opens. One for info and the other for detailed options. In the detailed options we could add all the options that taiga has such as manually defining a folder for the anime.

[...]

PR #191, opened 2016-03-01
[Qt][libmal] Tags support

Birdulon:
Adds a column for the tags string, adds a button to the side panel to edit show tags.


Birdulon:
Made it a bit more useful with the addition of a filter.

Birdulon:
With the ability to now filter for tags specifically and ignoring order, I now consider the functionality to no longer be rudimentary. Retaining column state and allowing users to change the order of columns should make the rest of #141 trivial, just a matter of adding the remaining columns and perhaps a dialog box to edit less-important data fields.
I'm considering adding the ability to click on a tag to add it to the filter (and perhaps the same functionality for all fields besides the title), but apart from that I can't think of anything else that would make the tag support nicer or more usable at this point.

Bonus screenshot of per-field filtering
trackmaadvancedfilter

z411:
Perfect, thank you. I'm especially liking the filter.

PR #203, opened 2016-04-22
Error escape play_random when no episodes are added to randomly pick from

### LINKED ISSUE #202: play_random gives IndexError when it can't find any new episodes.
Trackma crashes when it can't find any unwatched episodes to randomly pick from. Even when there should be episodes available to pick from for a single show.

Traceback:

```
  File "/usr/lib/python2.7/site-packages/trackma/ui/qtui.py", line 3011, in run
    ret = self.function(*self.args,**self.kwargs)
  File "/usr/lib/python2.7/site-packages/trackma/ui/qtui.py", line 2931, in _play_random
    (show, ep) = self.engine.play_random()
  File "/usr/lib/python2.7/site-packages/trackma/engine.py", line 708, in play_random
    show = random.choice(newep)
  File "/usr/lib/python2.7/random.py", line 275, in choice
    return seq[int(self.random() * len(seq))]  # raises IndexError if seq is empty
IndexError: list index out of range
```

Interestingly, as mentioned in the comment, `newep` seems to be empty in this case but it should have found unwatched episodes of one show which _are_ picked up by 'Play Next' and library scan, but seem to be ignored by play_random.  When new episodes are available for multiple shows, it does work, and will (randomly)  also select episodes of the first show. 

I don't know why this is, but it seems to me that, even if there is only one show to pick from it should still "randomly" select that one show, instead of ignoring it completely.

ToostInc:
Catches the error gracefully instead of crashing out completely.

Somewhat fixes #202.

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -700,10 +700,16 @@ class Engine:
         library = self.library()
         newep = []
 
+        self.msg.info(self.name, 'Looking for random episode.')
+
         for showid, eps in library.iteritems():
             show = self.get_show_info(showid)
             if show['my_progress'] + 1 in eps.keys():
                 newep.append(show)
+
+
+        if not newep:
+            raise utils.EngineError('No new episodes found to pick from.')
 
         show = random.choice(newep)
         ep = self.play_episode(show)


z411:
Thank you.

PR #217, opened 2016-07-28
Py3 CLI

mutilx9:
The python3 port broke the CLI UI because some strings are encoded to utf8. In python2, that changed str to unicode to show utf8 characters, but in python3 it changes str to bytes; and str is the type wanted in this case. The other UIs worked with almost all the strings being bytes, but the cli breaks when trying to concatenate str with bytes implicitly.

Also, I changed the complete\_filter method because the Cmd complete\_\* methods should return a list.

I checked every cli method and I think all worked as intended. The cache files should be deleted before trying it. **I didn't check if the melative change is needed** because I don't have an account, please someone check it out before merging.

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -289,7 +289,7 @@ class Engine:
         showdict = self.data_handler.get()
         # Do title lookup, slower
         for k, show in showdict.items():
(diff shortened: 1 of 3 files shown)


z411:
Yes, that's good. Thank you.

PR #227, opened 2016-08-07
[Library] Inotify and multiepisode support

Birdulon:
Inotify trackers now add episodes to library when detected (via creation, moving into watched directory or even just general access), and remove episodes from library when deleted or moved out of the watch directory.

Improved multiepisode behaviour of library (no more gaps in Momokuri bar!).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ NAME = "Trackma"
 REQUIREMENTS = []
 EXTRA_REQUIREMENTS = {
     'curses': ['urwid'],
-    'GTK' : ['PyGTK', 'Pillow'],
+    'GTK': ['gi', 'Pillow'],
     'Qt': ['Pillow'],
 }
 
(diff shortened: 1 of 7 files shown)


z411:
Thank you. I was planning to have Play Next use the library instead of searching the directory for the episode every time so this is useful.
I was wondering what to do in the case inotify wasn't available. Maybe have an option (enabled by default) to re-scan the library at start?

Birdulon:
I was planning on expanding the Rescan Library behaviour in Qt (adding a dialog that asks whether to scan for the current tab or full list, whether to skip the cache), so I can look into adding those two things to Qt along with that. I might have a go at curses and gtk afterwards.

z411:
Thank you.
By the way, the call to scan_library at start should definitely be handled by the engine, not the UI, as you'd have to redundantly implement the same thing in every interface this way. Pushed a commit to reflect this change; I also changed the config value to `library_scan` instead as to be more self-explanatory. Or maybe `library_scan_at_start`, as `library_scan` sounds more like what the tracker does with inotify.

Birdulon:
Cheers. I'll put up another PR once I make the Rescan dialog.

PR #229, opened 2016-08-24
Fix KeyError due to completed manga status

ToostInc:
Changes the status translation of completed manga in the Anilist library to reflect changes (?) to the API.

Fixes trackma crashing due to KeyError when using Anilist.

### DIFF
--- a/trackma/lib/libanilist.py
+++ b/trackma/lib/libanilist.py
@@ -106,7 +106,7 @@ class libanilist(lib):
             self.airing_str = "publishing_status"
             self.status_translate = {
                 'publishing': utils.STATUS_AIRING,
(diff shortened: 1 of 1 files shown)


joshstar:
Hmm I didn't notice this actually changed (it shouldn't have), sorry about that. Guess I'll keep it that now most of the manga app's using the anilist api have updated for the change.

ToostInc:
Yeah, I couldn't find out either whether it recently changed or had been broken for a long time, I looked at your commits on the documentation repo and though the .rst file I checked the status types against was brand new they didn't mention anything like that.

Might be this just snuck under the radar somewhere a long while ago because I imagine the anime tracking function is much more widely used and people have to actively switch mediatypes to find it.

PR #231, opened 2016-08-27
setup.py: fix requirements issue with `gi` package.

### LINKED ISSUE #228: setuptools gui_scripts cannot find gi package.
a very helpful user of the AUR package noticed trackma-gtk won't launch on Arch Linux, giving the following traceback:

```python
  File "/usr/bin/trackma-gtk", line 9, in <module>
    load_entry_point('Trackma==0.6.1', 'gui_scripts', 'trackma-gtk')()
  File "/usr/lib/python3.5/site-packages/pkg_resources/__init__.py", line 542, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2575, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2234, in load
    self.require(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2251, in require
    items = working_set.resolve(reqs, env, installer)
  File "/usr/lib/python3.5/site-packages/pkg_resources/__init__.py", line 829, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'gi' distribution was not found and is required by the application
```

The problem seems to be that since 18ca09c, which adds `gi` as a requirement, setuptools can't find the gi package, despite it being installed (under `python-gobject` on Arch).
Removing this requirement, either in setup.py or the resulting egg, "fixes" the issue and trackma-gtk will simply run if you have python-gobject installed.

The odd thing is that `setuptools.find_packages()` *does* manage to find the gi package and it's related parts when run manually. I did find [this](http://stackoverflow.com/questions/21581799/how-does-one-declare-a-dependency-on-gi-repository-in-setup-py-and-required-c-l) as well, but admittedly, I have no clue when it comes to gi so I don't know whether this is even relevant or true.

Anyway, this looks like a setuptools problem to me, but I figured I'd bring it here first anyway, see if people are having this problem outside of Arch. 
If it's just Arch, then that is easily fixed in packaging by simply removing gi from the setuptools requirements, since the package itself has all dependency information anyway.

ToostInc:
refer to  `pygobject` in the GTK requirements, so that setuptools can find it.

Fixes #228; tested on Arch Linux, but will probably work the same on other distributions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ NAME = "Trackma"
 REQUIREMENTS = []
 EXTRA_REQUIREMENTS = {
     'curses': ['urwid'],
-    'GTK': ['gi', 'Pillow'],
+    'GTK': ['pygobject', 'Pillow'],
     'Qt': ['Pillow'],
 }
 


z411:
Thank you.

PR #236, opened 2016-09-04
[libanilist] Proper error on no search results

Birdulon:
The Anilist API now seems to return ["error"] when there are no search results, which causes a problem when the parser then tries to take a string literal index of the string. Added a check for non-dict items to avoid this.


z411:
Fixed by #238

PR #239, opened 2016-10-01
add error message for openContainingFolder() if no folder was found

MadeOfMagicAndWires:
in case engine.search_video() can't find anything, there is no folder to open, so let the user know.


MadeOfMagicAndWires:
Sorry this was supposed to go somewhere else, but Github is getting too smart these days.

PR #208, opened 2016-05-04
open containing folder (gtk ui)

anjandev:
I have started working on an open containing folder function as discussed in #133 . Im having trouble with getting the `task_openContianingFolder` function to work and would like to get some help with it. Sorry if I overlooked something simple or if my approach is completely wrong.

Issue #133, opened 2015-08-08
Open containing folder for anime / Deleting anime


anjandev:
It would be nice to add an option to ask the user if they want to delete the entire anime's folder from their hard drive when the show is right clicked or when the show is deleted from the list. It would allow one to keep their anime folder under control and use less space. It's hard to look through a giant folder with files everywhere.

ToostInc:
Isn't that more a job for the user to keep his folders organized though?
Similarly, this organization will really differ per person, and will be hard to get working safely. Deleting files based on a regexp pattern seems very unsafe.

anjandev:
I mean updating hummingbird/MAL is a job for the user but we have taiga automate that cause we are lazy. I think opening thunar/nautilus/ranger and asking to confirm if that was the right folder would be a nice remedy to the problem of having a mistake. There's not many ways to organize anime. Most people do it like ./anime/series_title

ToostInc:
Or everything in a large map called anime, or an even larger map(/partition) called media. Or a folder for every episode... 

It's best to let the deletion of files up to the user really, but if https://github.com/z411/trackma/issues/117 becomes a thing, you can use that to delete the files.

dyskette:
> It would be nice to add an option to ask the user if they want to delete the entire anime's folder from their hard drive when the show is right clicked

I think this is out of the scope of trackma. Altering local files should be left to their respective tools. 

Trackma is handling online lists, that are not tied to local files. You could be syncing between different machines, have external hard drives well organized or everything in the "Downloads" folder.

> or when the show is deleted from the list.

Is this usecase often?

z411:
Indeed; besides what's been said, I wouldn't want Trackma to go around the user's hard disk deleting files...

anjandev:
When I right click an anime, could you add an option in the gui to view containing folder? When clicked, this would open the folder in which trackma thinks the show is stored. The user then can delete the show manually through the file manager of their choice.

z411:
> When clicked, this would open the folder in which trackma thinks the show is stored
How could Taiga guess this? What about the location of the latest episode?

anjandev:
I assume thats it. If the show has never been played before, trackma could use the search engine to find the episode #1's directory and assume that that is where all the episodes are stored. 

I am assuming taiga makes a database for where the shows' folders are. If the folder no longer exists where the database says it is, taiga searches for the show in the folders and updates the database.

anjandev:
I think this issue can be reopened.

[...]

PR #240, opened 2016-10-01
open containing folder (gtk ui)

anjandev:
Added open containing folder button to gtk ui as discussed in #133. The button opens the folder containing the anime with whatever file manager the user has as default and displays an error if no folder is found.

As @MadeOfMagicAndWires recommended, I opened a new PR (and closed #208) to keep the git commit logs clean. Special thanks to @MadeOfMagicAndWires for helping me out. I will try working on adding this function to the qt and cli uis.


anjandev:
@z411 what do you think of this?

z411:
Looks fine, although I'd do 2 things before merging:

1. The library is in an usable state now, so it'd be much faster to use that instead of searching for the episode path each time, since the library already has the full paths of all available episodes on disk. I'll migrate the engine.play_episode routine to use the library too, and hopefully deprecate the engine._search_video routine altogether.
2. Is it necessary to use a separate thread for this?

z411:
Done (https://github.com/z411/trackma/commit/bf2d5aa7dc1a86735424aa0a27c2615d7361d788).

Now you should be able to use the engine.get_episode_path function to get the full path of the episode from the library. So just something like:

    filename = self.get_episode_path(show, 1)

should be enough.

MadeOfMagicAndWires:
I'll work on it and make another clean pull with the conflicts resolved as well.

(Really wish Github could handle a pull request using rebase instead of merge all the time...)

z411:
Superseded by #247

PR #247, opened 2016-10-14
open containing folder (gtk ui)

MadeOfMagicAndWires:
Alright, here it is

The original idea and pull are of course by @anjandev, with the added comments from https://github.com/z411/trackma/pull/240#issuecomment-253723565 and I also greyed out the 'Open containing folder' menu button if no show is selected.

@anjandev: Sorry your commits got a bit mangled, 3-way diffs are not my strong suit. In the end things would have probably been easier if I just asked to be a collaborator on your fork repo, but you live and learn.


z411:
Looks good; although the "task" functions are reserved for those that work inside a new thread. Could you move it all to `do_containingFolder` instead? That and we'd be done.

MadeOfMagicAndWires:
Done.

I did just realise that, although it's more usable than directly calling nautilus, this is of course in no way cross-platform. 

This being Gtk, I don't think we have to worry about it being used much on other systems but we could also add an OS check and launch `finder` or `explorer.exe` on MacOS or Windows?

MadeOfMagicAndWires:
I've got a commit for os specific commands ready, but it probably still needs some testing and so I don't know if I should add it to this pull, I'll let you decide.

z411:
Let's use a separate request for that.
Will merge this, thanks.

PR #254, opened 2016-11-15
[Tracker][Engine][Qt] Prompt to add unrecognised shows.

Birdulon:
Adds an option to prompt to add new shows to your list as/after you watch them. The prompt is the same as the regular search/add from remote dialog, however it starts with the show title filled in (not searched without user input to prevent privacy leaks) and closes after a show is successfully added, then updates the episode counter to match the one that was detected.

Known issue: Changing tracker options requires a restart to take effect. Would it be overkill to add a tracker restart to settings changes that involve the tracker?

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -67,7 +67,8 @@ class Engine:
                 'sync_complete':     None,
                 'queue_changed':     None,
                 'playing':           None,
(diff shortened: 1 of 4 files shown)


z411:
Good. Thank you.

PR #260, opened 2016-11-29
[Qt] Split setGeometry into resize and move. (fixes #246)

### LINKED ISSUE #246: [win32, Qt] Remember window size and position not remembering position
I've noticed a drift when opening trackma-qt multiple times, in which it slowly creeps up the screen with each time it opens.  Eventually, this gets bad enough that the user can't grab the titlebar, and then it can even get bad enough that the menus can't be opened.  I have to resort to either A) remembering to drag the window down to the middle of the screen before the title bar gets swallowed by the edge of the desktop, or to go into ui-qt.json in  %USERPROFILE%\.trackma and forcibly reset last_y to something well within the screen boundaries.

Example: I force last_x to 282 and last_y to 221.  Launch it once, it's in more-or-less the right place.  But after closing it, last_x is now 274 and last_y is 190.

It seems to be fine with remembering the window size, so this is just positioning.

It'd actually be nice to have an 'Always open window in the center of the screen' option.

Birdulon:
Workaround taken from the bottom of https://www.tbi.univie.ac.at/~pmg/tutorials/QT/html/geometry.html
I'm disgusted by the need for it, but it seems to fix the problem on my window manager setup.

EDIT: If someone on Windows could test this it would be much appreciated!

### DIFF
--- a/trackma/ui/qtui.py
+++ b/trackma/ui/qtui.py
@@ -374,8 +374,8 @@ class Trackma(QMainWindow):
 
         #self.setMinimumSize(740, 480)
         if self.config['remember_geometry']:
-            self.setGeometry(self.config['last_x'], self.config['last_y'],
-                             self.config['last_width'], self.config['last_height'])
+            self.resize(self.config['last_width'], self.config['last_height'])
+            self.move(self.config['last_x'], self.config['last_y'])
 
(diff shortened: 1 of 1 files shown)


z411:
Thank you.

PR #262, opened 2016-12-03
[Tracker][Qt] Add option to ignore files outside of media directory

Birdulon:
With the addition of prompting to add shows not in the list, the tracker responding to media outside the watch directory (i.e. when watched folders are reloaded while a player is running) is a nuisance.
Optional with default to off as it may interfere with users who have symlinked watch directories.


Birdulon:
Any comments?

z411:
Sorry I've been away from home which is why I also haven't been able to finish the Kitsu implementation, I'll review as soon as possible.

z411:
Actually this was the original *and* default behavior, adding the searchdir as a parameter to the lsof command should be enough to fix this. Also even if this was configurable, it wouldn't normally work with inotify as they watch the searchdir only.

Birdulon:
Ah ok, I'll rewrite that bit to add the searchdir as a parameter instead of matching the full filename afterwards. If it does what I think it does it may also solve an issue where multiple instances of a media player are open and the one playing something else is used for detection purposes, causing the anime episode to get through unnoticed.

The issue with inotify is that events in the watch directory can be generated by e.g. file browsers scouring for .directory files, which in turn trigger detection of whatever the media player is currently playing. Not an issue when anything off the lists are ignored, but when they are prompted to be added it can quickly become a nuisance.

z411:
>I'll rewrite that bit to add the searchdir as a parameter instead of matching the full filename afterwards.
Yes, but my main point is, should this really be configurable? The media tracker reacting to things outside of the media directory sounded like a bug to me, so I'm not sure if people would actually want this behavior.

>If it does what I think it does it may also solve an issue where multiple instances of a media player are open and the one playing something else is used for detection purposes, causing the anime episode to get through unnoticed.
Now that you mention it I can see this happening; I haven't tested it yet.

>The issue with inotify is that events in the watch directory can be generated by e.g. file browsers scouring for .directory files, which in turn trigger detection of whatever the media player is currently playing.
Maybe we could make the extension regex (tracker.py:125) global, and use it to filter inotify event names (event.name in pyinotify's case); this way only events from video files would trigger polls to lsof.

z411:
This got resolved by inotify and lsof shouldn't be doing it. If it's happening then it's a bug. Not sure what to do in MPRIS case, though.

FichteFoll:
mpv includes a `xesam:url` key in the metadata map that has the url-percent-encoded path to the playing file, next to `xesam:title`, so that could be used. Vlc does as well. Don't know about other players (because I don't have them installed).

z411:
Interesting, that definitely could be done. Thanks for the pointer.

PR #270, opened 2017-01-05
Various things

Hairo:
- Add missing function in the plex tracker (lost in the refactoring)
- Put an item counter to the curses filters


z411:
Thank you.

Issue #181, opened 2015-12-02
Rescanning library for plan to watch 


hbers:
Rescanning library used to find anime in every category but now it only detects the ones in the watching category. When I first noticed it I thought that it was intentional. If that's not the case can you add an option to scan for all categories?

z411:
It is indeed intentional, since rescanning the whole list, especially if you have a lot of completed shows, takes more time, and I assumed that it shouldn't be necessary to scan for shows in the Completed list. Maybe scan for all shows except completed (since it tends to be the largest)?

hbers:
Yeah that's more practical definitely. Additionally I think there is no need for scanning Dropped too.

Fasolt:
I'd like to second this request, as it'd helpful to be able to just pick an anime from my PTW and click on Play Next, with everything working automatically. Perhaps it would be possible to add an option in config.json for choosing which list sections the user would like scanned; that way, people who don't want the broadened library scan won't suffer a performance drop.
Otherwise, simply scanning Watching, PTW and possibly On Hold would be sufficient, I imagine.

MadeOfMagicAndWires:
I have roughly implemented this, optimising out Dropped however would require more rigorous optimization however, so I'd like some user input on how big people's Dropped lists tend to be (and if you keep the files) to see if that's worth it before filing a PR.

I'd probably recommend keeping this setting off regardless though, since a library scan went from just over 2 seconds to ~30+, and that is *without* Completed shows.

hbers:
Why not just scan the PTW and Watching then? There's no need to scan for Dropped and Watched in my opinion. It's not like people don't know what they've watched and what is on their drives.

z411:
He did say
> and that is without Completed shows
which is weird, unless you have a lot of shows on hold/dropped/PTW.

MadeOfMagicAndWires:
I have 11 shows for On-hold (no files of dropped series, for obvious reasons) not sure if that's a lot, but the main killer for me is there's quite a lot files ( several series with 50+ episodes) and some large files as well in between those shows that take quite a bit of time to process. 

But that's just one case, so I'm trying to figure out if optimizing Dropped out dropped is even worth it. 

The problem with scanning for PTW and Watching as @hbers suggests is that the current data model doesn't keep track of what status is PTW, so we can't specifically scan for it. 
The current model only keeps track of "Watching" and "Completed" as specific statuses because we can be fairly certain those statuses will exist across all services, so currently I scan for everything *except* completed.

[...]

PR #277, opened 2017-01-29
Scan library for multiple statuses

MadeOfMagicAndWires:
Implements #181.

To recap this PR adds the following:
+ The engine now optionally scans the mediafolder for series from all non-completed statuses (PTW/On-    Hold/Dropped)
+ Added a setting that en- or disables this feature (disabled by default)
+ Added corresponding settings to the GUIs (including a missing 'scan lib at startup for GTK)
+ The GUI remains operable whilst scanning

The pros and cons of this feature have been discussed, but it's probably fine as it is? I made it an optional feature to begin with, it's there for those that want it, and if it's too much of a drain it can be turned off.

Since I've had little feedback I figured I'd file a PR anyway and let you decide whether to merge it now or if it still needs work

Possible things to improve:

- [ ] I could fairly easily further optimise things to only scan for Watching/PTW, but I didn't want to mess  with the datamodel for a single feature just yet. let me know if you think this is worth it.

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -181,9 +181,9 @@ class Engine:
 
     def _get_tracker_list(self, filter_num=None):
         tracker_list = []
-        if type(filter_num) == None:
+        if isinstance(fliter_num, None):
             source_list = self.get_list()
-        elif type(filter_num) == list:
+        elif isinstance(filter_num, list):
             source_list = []
             for status in filter_num:
                 if status is not self.mediainfo['status_finish']:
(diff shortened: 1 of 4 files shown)


guicale:
Is there any reason this pull request hasn't been accepted?

z411:
I was thinking of having settings tab for site-dependent options, and have an option to check which statuses you want checked. That would fix this in a cleaner way, and also make it possible to do things like scanning only Watching and PTW, but on the other hand it makes text-based configuration harder (for curses and CLI users) so I think I'll just merge this.

PR #279, opened 2017-02-15
fix 'Using Plex' spam introduced in the refactoring

Hairo:


z411:
I don't use Plex but why is the second "Using plex" message (the one in the while loop) necessary? Shouldn't the first one be enough?

Hairo:
It's there in case of connection issues, it's displayed when it reconnects back to the plex server, though i could change it to say "Reconnected to plex" or something.

z411:
That would be better. Just that and I'll merge.

z411:
Thanks.

PR #287, opened 2017-03-26
Fix the file paths in the lastest plex update

Hairo:
As of the lastest plex update (1.4.4.3495) it no longer returns the full path of the videos in the usual url (http://host:port/status/sessions), apparently it only returns it for the non-transcoded files, this commit makes the tracker read filepath from somewhere else (http://host:port/library/metadata/id), thankfully this still works in the previous versions.


z411:
Thank you.

PR #293, opened 2017-04-10
Show names are based on the relative path

manulaiko:
Now show names are based on the relative path of the file to the search directory from the configuration file instead of just the file name.
This way `/home/manulaiko/Documents/Animes/Bakemonogatari/01.mkv` with `/home/manulaiko/Documents/Animes` as search directory will count as `Bakemonogatari 01.mkv` instead of just `01.mkv` and you'll be able to use the `Play next` button.


ZLima12:
So uhh... is there any intention on merging this?

PR #303, opened 2017-07-27
Fix typo, fix isinstance() call

matoro:
Not sure how this made it into master but it totally borks the whole thing for me since latest pull.  This *may* be because I'm on python 3.6.2, don't have time to test on other versions though.  Anyhow, this fixes it for me, or so it seems.


z411:
Thank you.

PR #326, opened 2018-02-08
curses: 256 color support

drewt:
Adds support for 256 colors on terminals that support it.  Also fixes an issue where the 'body' display attributes aren't applied to blank space after the show list or on the details view (I'm assuming this was the intent of the 'body' attributes; if not I can add another set of attributes for this purpose).


z411:
Great, thank you.

PR #329, opened 2018-03-04
Fix plex tracker crashes

llwu:
This is just because i was getting a bunch of crashes from the plex tracker.

Some exceptions due to each `observe()` not being atomic (i.e. the status can go `NOT_RUNNING` after it was last checked but before the xml info is needed).

Also a queue atomicity issue.

### DIFF
--- a/trackma/data.py
+++ b/trackma/data.py
@@ -371,8 +371,12 @@ class Data():
 
             # Run through queue
             items_processed = []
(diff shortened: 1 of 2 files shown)


z411:
Thank you. Is there a way to change this to catch the expected exceptions (as opposed to `except Exception`)? Same with your other PR.

llwu:
Addressed

PR #330, opened 2018-03-04
Forge filename in Plex tracker.

llllvvuu:
There is no need to parse the original filename (this sometimes fails) if Plex already has the title and ep #.


Hairo:
The `grandparentTitle` tag seems to be used only in newer versions, also the method used now was a fix for a previous version that only returned the filename when the file was not being transcoded (see #287).

> this doesn't always work

when does it not work?

llwu:
Thanks for pointing this out. The only reason I wanted this is because I have some strange filenames such as `Kill.la.Kill.S01E01.1080p-Hi10p.BluRay.FLAC2.0.x264-CTR.[98AA9B1C].mkv` which trackma can't parse for the title and episode. But Plex has that information.

Hairo:
Oh, the issue seems to be with [AnimeInfoExtractor.py](https://github.com/z411/trackma/blob/master/trackma/extras/AnimeInfoExtractor.py) then, which is what trackma uses to get the info from filenames, getting the info from the plex agents is not a good idea either because they (depending on which one is used) sometimes label the shows incorrectly (either by labeling them as the wrong season, not even recognizing OVAs as separate entries, or straight up getting the title wrong). Either way this is the first time i see that way of naming anime episodes.

PR #346, opened 2018-06-04
Anilist v2 (#343)

### LINKED ISSUE #343: AniList API v1 shutdown - Migration to v2
AniList API v1 was given 6 months before it was meant to be shut down after API v2 went stable.
I made a few announcements about this and sent out emails to developers who had registered API v1 clients.

It looks like trackma is still using API v1, which was meant to be shut down on May 1st, due to some issues on AniList's part, API v1 is still running, though will likely only up for another 2 weeks or so, as the majority of applications have already migrated to v2.

Super sorry if you didn't hear about all this, as this would be super short notice.

[Api v2 docs can be found here](https://github.com/AniList/ApiV2-GraphQL-Docs), if you check the migration folder theres a number of queries that aim to replicate v1's rest call data set, so hopefully they help with the possible migration.

Birdulon:
Had a bit more time to finish off my implementation. Feel free to close it if you've made significant progress with yours, I needed to learn the new API for my own reasons anyway.
Also if you decide to merge, feel free to change the client ID and secret pair to something under your account. (fixes #343)

### DIFF
--- a/trackma/lib/libanilist.py
+++ b/trackma/lib/libanilist.py
@@ -49,13 +49,13 @@ class libanilist(lib):
         'date_next_ep': True,
         'status_start': 'watching',
         'status_finish': 'completed',
-        'statuses':  ['Watching', 'Completed', 'Paused', 'Dropped', 'Planning'],
+        'statuses':  ['CURRENT', 'COMPLETED', 'PAUSED', 'DROPPED', 'PLANNING'],
         'statuses_dict': {
-            'Watching': 'Watching',
-            'Completed': 'Completed',
-            'Paused': 'On Hold',
-            'Dropped': 'Dropped',
-            'Planning': 'Plan to Watch'
+            'CURRENT': 'Watching',
+            'COMPLETED': 'Completed',
+            'PAUSED': 'On Hold',
+            'DROPPED': 'Dropped',
+            'PLANNING': 'Plan to Watch'
         },
         'score_max': 100,
         'score_step': 1,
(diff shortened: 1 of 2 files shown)


z411:
Thank you. I had already modified the basic functions but hadn't implemented the calls yet so I will review this and merge it with what I had done.

Also a concern of mine is the token; wouldn't the token refresh fail since the authorization code expires after its use?

z411:
Thank you. I had already modified the basic functions but hadn't implemented the calls yet so I will review this and merge it with what I had done.

Also a concern of mine is the token; wouldn't the token refresh fail since the authorization code expires after its use?

trackma/lib/libanilist.py
```diff
@@ -70,13 +70,13 @@ class libanilist(lib):
...
-            'reading': 'Reading',
-            'completed': 'Completed',
-            'on-hold': 'On Hold',
-            'dropped': 'Dropped',
-            'plan to read': 'Plan to Read'
+            'CURRENT': 'Watching',
+            'COMPLETED': 'Completed',
+            'PAUSED': 'On Hold',
+            'DROPPED': 'Dropped',
+            'PLANNING': 'Plan to Ratch'
```
FichteFoll: `Watch` typo here.

Birdulon: Good catch!

Birdulon:
Re: Token - I'm not entirely sure what's going on with it, but my user.json does have different access_token and refresh_tokens. Token expiration is a year away so unsure how to otherwise test that.

z411:
Merged most of your changes into the `anilistv2` branch, thank you.

PR #357, opened 2018-06-21
[GTK] Display lists nicely in detail window

SynderBlack:
also fixes lists having empty strings in them
for example:
`['action','adventure','','','']` -> `action, adventure`


SynderBlack:
In relation to this PR, this happens in the other UIs as well, should I add those changes too?

PR #364, opened 2018-08-24
[QT, GTK] fixes for when PIL is unavailable

SynderBlack:
Fixes crashing for QT and an error from GTK when you don't have PIL installed and try to load a thumbnail.


z411:
Thank you.

PR #367, opened 2018-08-26
[GTK] add command line arg to start in tray

doron-goldstein:
Note: this can cause an inconsistency in the settings, as this sets the settings manually.
This can be solved by replacing this with a `-c` option or such, to instead use a different config file. But I think having an option dedicated to starting in the BG is much more useful.
Another solution would be to just manually start hidden but that would be inconsistent with the settings internally, which can be all kinds of not fun.

TL;DR, looking for some opinions because this isn't the best impl.


ahmubashshir:
> Note: this can cause an inconsistency in the settings, as this sets the settings manually.
> This can be solved by replacing this with a `-c` option or such, to instead use a different config file. But I think having an option dedicated to starting in the BG is much more useful.
> Another solution would be to just manually start hidden but that would be inconsistent with the settings internally, which can be all kinds of not fun.
> 
> TL;DR, looking for some opinions because this isn't the best impl.

It already implemented in a323b107

Issue #190, opened 2016-02-21
Any plan to translations? 


AlejandroC90:
I wonder if you're planning to add translations in the future, I could help with the Spanish one.

dyskette:
See #149

z411:
Yes, this is a duplicate of #149 -

PR #370, opened 2018-10-21
Make Guis translatable

### LINKED ISSUE #149: Make the GUIs translatable
I don't know if that's planned, but I think it should be great to make the GTKs and Qt GUI translatable.

I would be very pleased to help in that and to translate Trackma in French :)

anjandev:
Fixes #149 and #190.

I started to specify translatable text for the cli ui first because I wanted to get a working prototype. Changing the locale option in your config to "es" will display "Adios!" instead of "Bye!" when you exit the program.

Please give me feedback on this implementation. I will look to specify the translatable text in the other uis if this is ok. 

We should prolly get one of those open source webguis where mutliple people can easily specify translations. I dont know any other languages other than english so I would not really be able to help translate.


anjandev:
@z411 Is there any reason this was never merged and is there anything I can do to get it into a state that can be merged?

FichteFoll:
Unfortunately, this PR is not mergeable by now. There appear to be two commits at the end of the branch that could be cherry-picked, but will most likely conflict with other changes from the past 4 years or so.

My personal experience with translations is very limited (read: none) and I cannot comment on how easy or hard it would be to integrate existing web translation tools or UIs into a Gitlab PR workflow (or whether we'd just export translation off an external web tool and put them into the source tree). 

Because of how stale the PR is, I will close it to clean up the list of open PRs, but it will still be referenced from #149.

PR #377, opened 2018-12-07
Gtk and Qt5 Flatpaks

thebitstick:
I've tested the Gtk Flatpak, Qt5 Flatpak untested.


lionirdeadman:
Master is probably not a good idea since it's a moving target.

thebitstick:
I tried changing the base to experimental but there was quite a difference in commits, would it be possible to create a flatpak branch?

lionirdeadman:
I meant for `org.gnome.Platform`, should've specified, haha.

thebitstick:
oh

PR #379, opened 2019-01-04
libkitsu: add missing user start & finish dates

uralica:
Hello friends,
The current kitsu behaviour is to fetch only the series "start" and "finish" dates, but not the user timetables. This small addition adds the latter. Please double check as this is my first commit. >_<


z411:
Thank you!

PR #383, opened 2019-03-17
Clean up trackers when unloading engines

### LINKED ISSUE #382: Trackers aren't cleaned up when Engines unload
When Engines unload they never call tracker.disable() which trackers to stick around indefinitely. This can cause duplicate trackers when switching between media types or accounts.

I noticed it because I had a very long running session on a machine where I never installed pyinotify so I had many polling trackers running at once.

awused:
Fixes #382.

Call tracker.disable() when unloading engines.

Check tracker active state before reacting to incoming events.

Remove unused tracker.enable() method.


z411:
Thank you.

PR #390, opened 2019-04-14
[WIP] Pylint fixes

InitialFail:
Trying to cleanup some code via Pylint. Most of the errors are gone, but there is still a huge list of warnings to work through. Also, I've made a few guesses about some changes here and there, and commented out code I didn't know what to do with.

Would pylint (or some other static analysis tooling) be something of interest to you? I don't think I'll have a lot of time to learn the code base and work through the majority of the problems.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,10 @@
+[MASTER]
+persistent=no
+jobs=4
+extension-pkg-whitelist=PyQt5
+disable=
+	invalid-name,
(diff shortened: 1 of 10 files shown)


z411:
I'm definitely interested, thank you. The parts you commented out as TODO are leftovers of the previous usage of QListWidget for the show list that got recently switched to QListView, they need to be worked on, I'll do that now.

PR #396, opened 2019-04-20
trackma.ui.qt.mainwindow: Fix old statuses_nums reference

InitialFail:


z411:
Oh I missed this. Thank you.

Issue #402, opened 2019-04-28
Timer still counts down even if paused + MPRIS


Soitora:
I started an episode earlier but two minutes in I had to go brb so I paused and didn't resume watching for several hours, yet it proceeded to update the episode as watched even though it had recognized (using `tracker` command) that it was paused.

z411:
Hi, that's expected behavior. There's no way for Trackma to know if the video is paused or not since it works by checking if a file is open. The way to actually implement this would be to, for example, write an MPRIS tracker, but your played would need to support that as well.

Soitora:
Hey this was an oversight on my part, I thought I remembered the `tracker` command showing my Plex tracking as paused when I paused but after re-checking it did not, maybe something that could be thrown in for the Plex users as it's very easy grabbing playing status?

Meybe you, @Hairo, could look into it if you ever got the spare time ? :)

z411:
I was actually thinking of throwing together an MPRIS tracker so I will eventually have to develop a pausing mechanism for the timer that could be used by the Plex tracker as well, yes.

Hairo:
While getting the playing state from a plex session would be trivial, i'm not sure how the tracker would handle that.

z411:
Just implemented a MPRIS tracker, and with that, a pause mechanism for the Tracker. It works with MPRIS as I tested it, and @Hairo you should be able to use it too from the Plex tracker. Just call `self.pause_timer()` and `self.resume_timer()` respectively.

Soitora:
Awesome! :smile:
I'll try it out when available

FichteFoll:
With the mpris checker, couldn't you also check for a % completion threshold of the played file using its metadata? That sounds like a more solid approach for shows of different length and I believe is also the preferred practice for scrobbling music (to lastfm, for example). Should I open a separate issue for that?

z411:
As in using the playing position and total length or just the length?

Soitora:
I think @FichteFoll means that if the show is **25 minutes** long and you have it set to mark as complete at **80%** then the program waits until it reaches **20 minutes** and then marks as watched. This could be a replacement or an alternative to the timer and would technically be more accurate.

z411:
Yes, but I see two ways to do this:

1. Timer: Calculate the 80% (20 minutes) and then countdown, independently of what the user does. You are forced to watch at least 20 minutes of the media for it to update.
2. Playing position: Update when the player reaches 80% of the video. This would mean that it would update if you seek the video to the end.

[...]

PR #399, opened 2019-04-21
Fixes #393 (only for claimed severs), and display a message when the sever is claimed. 

Hairo:
Fixes #393, only for claimed servers because it's the only way to get user info (to identify who is watching it), only tracks if the login username and the user watching the video returned by plex match.

Also, display a different message when the server is claimed and trackma can't access it (because login info is needed).

PR #405, opened 2019-04-29
Plex: Check for the user in both observe() and the xml info and Implement pause/playing functionality

Hairo:
For #402 and #399.


z411:
Thank you!

PR #415, opened 2019-05-05
Gtk: Fix tray interactions with main window

dyskette:
Because I broke the tray with the refactor, now I'm fixing it again. The next one in the list is to restore keybindings.


z411:
Thank you as always.

PR #426, opened 2019-05-16
Use attributes.ratingTwenty in libkitsu

ahmubashshir:
Current kitsu api provides rating and ratingTwenty for rating.
As rating is now deprecated we should use ratingTwenty.


z411:
I think I did this before... but it seems I never pushed it. Thank you!

PR #429, opened 2019-05-20
Added screenshot for Gtk3.

ahmubashshir:

PR #430, opened 2019-05-20
Fix gtk papercuts

dyskette:
This is a batch of commits related to fixing all these little mistakes and errors that were left after the refactor. 

- Restoring keybindings.
- Add shortcuts window.
- Using GtkApplication.
- Restore column toggle functionality.
- Refactors image download thread class.

With this, I think the UI is mostly done. Now it will be easier to experiment with and implement the redesign, but like I said in #429 it will take me a while.


z411:
Thank you, very appreciated.

PR #428, opened 2019-05-20
Added Variables for mediatypes from kitsu

ahmubashshir:
Added:
	Manga,Manhua,Novel,Oneshot,...
	GTK:Show mediatype in search.

PR #437, opened 2019-05-24
Removed unrelated commits from #428 and Opened this.

ahmubashshir:


z411:
Mhm... Shouldn't we maybe move these to the APIs instead of `utils`? In the future the APIs might need other media types and it'd be a pain to modify utils.py every time there's a new media type that's not supported. What about just returning it as a `str`, or having the list of types in the `mediatypes` dict?

ahmubashshir:
We shouldn't add them to mediatypes dict as they are subtype.
And dict returned by api doesn't have a subtype key.

z411:
> We shouldn't add them to mediatypes dict as they are subtype.
> And dict returned by api doesn't have a subtype key.

Could you clarify? The `subtype` field in kitsu is equivalent to the `type` field in the show dictionaries.
For the record, what I meant is moving `utils.TYPE` to:

https://github.com/z411/trackma/blob/ecf17e99c18d657dc5ec39f1437e1e32d4bd2abf/trackma/lib/libkitsu.py#L67

It makes sense to me that the API provides this information because:

1. Types are API-depenent (Anilist has NOVEL as a manga type while Kitsu has it as a separate mediatype, VNDB has just completely different types altogether, etc.) so it makes sense to me for the API to provide that information and not to have it hardcoded into Trackma.
2. Types are mediatype-dependent. For example Anime has TV, OVA, MOVIE, etc. and Manga has DOUJIN, MANGA, NOVEL, MANHWA, etc.

ahmubashshir:
ok.I will try.

ahmubashshir:
> Mhm... Shouldn't we maybe move these to the APIs instead of `utils`? In the future the APIs might need other media types and it'd be a pain to modify utils.py every time there's a new media type that's not supported. What about just returning it as a `str`, or having the list of types in the `mediatypes` dict?

I'm closing this and opening #480

PR #439, opened 2019-06-20
Handle urllib.error.URLError like urllib.request.HTTPError

InitialFail:
Fixes issues where trackma crashes if it cannot establish the HTTP connections with the API endpoints.


z411:
Thank you.

PR #459, opened 2019-10-24
Qt: Make settings pages scrollable

amhndu:
Settings page would overflow and take the buttons out of the screen/dock
on screens with smaller resolutions (eg. 1366x768).
This will add a scroll bar if the pages would overflow.


trackma/ui/qt/settings.py
```diff
@@ -356,10 +356,12 @@ def __init__(self, parent, worker, config, configfile):
 
         # Content
         self.contents = QStackedWidget()
-        self.contents.addWidget(page_media)
-        self.contents.addWidget(page_sync)
-        self.contents.addWidget(page_ui)
-        self.contents.addWidget(page_theme)
+        for page in (page_media, page_sync, page_ui, page_theme):
+            scrollable_page = QScrollArea()
+            scrollable_page.setWidgetResizable(True);
```
FichteFoll: There's an extraneous semicolon here.

amhndu: The semi-colon muscle memory strikes again!

Thanks for reviewing, I've removed it now.

PR #468, opened 2019-12-30
Qt: use `exclusionPolicy` instead of `exclusive`

molkoback:
Changed in Qt 5.14: https://doc.qt.io/qt-5/qactiongroup.html#ExclusionPolicy-enum


lastweakness:
It does fix the issue. But is it backwards compatible?

molkoback:
> It does fix the issue. But is it backwards compatible?
It's not. There's also [setExclusive](https://doc.qt.io/qt-5/qactiongroup.html#setExclusive) method but I didn't notice it before.

Issue #413, opened 2019-05-02
Misc recognition issues


FichteFoll:
```
[D] Engine: Adding to library: /data/Video/Anime/data_tmp_Anime/[PAS] Houseki no Kuni - 05 [WEB 720p E-AC-3] [F671AE53].mkv
[D] Engine: Redirected to Houseki no Kuni 3

[D] Engine: Adding to library: /data/Video/Anime/The Tatami Galaxy/[Opportunity] The Tatami Galaxy - NCED [BD 720p] [903A3AD6].mkv
[D] Engine: Redirected to Yojouhan Shinwa Taikei 1

[D] Engine: Adding to library: /data/Video/Anime/The Tatami Galaxy/[Opportunity] The Tatami Galaxy 10 - The 4.5-Tatami Idealogue [BD 720p] [FF757616].mkv
[D] Engine: Redirected to Yojouhan Shinwa Taikei 4
[D] Engine: Not a show, skipping: /data/Video/Anime/The Tatami Galaxy/[Opportunity] The Tatami Galaxy 11 - The End of the 4.5-Tatami Age [BD 720p] [6E03B068].mkv
```

These are the most obvious ones in my current library. The others stem from either the appropriate show now being on my list or with the completed status, which results in specials being recognized as episodes for the main TV show or vice versa. Things containing "NCED" or "NCOP" should probably not be recognized at all, if possible.

FichteFoll:
Enabled a whole scan now (with completed and dropped), but got mixed results. GuP OVAs are all recognized as the single-ep Anzio OVA instead of the "Girls und Panzer Specials" (as they are called on anilist, duh). And the file that actually contains the Anzio OVA is then recognized as the movie.

```
[D] Engine: Adding to library: /data/Video/Anime/[-__-'] Girls und Panzer [BD 1080p]/[-__-'] Girls und Panzer OVA 6 [BD 1080p FLAC] [B13C83A0].mkv
[D] Engine: Redirected to Girls und Panzer: Kore ga Hontou no Anzio-sen Desu! 6

[D] Engine: Adding to library: /data/Video/Anime/[-__-'] Girls und Panzer [BD 1080p]/[-__-'] Girls und Panzer OVA Anzio-sen [BD 1080p FLAC] [231FDA45].mkv
[D] Engine: Redirected to Girls und Panzer der Film 1

[...]

PR #473, opened 2020-01-21
Initial commit of Anitopy integration changes

MisterMcDuck:
Hello,

I've modified trackma to also leverage the anitopy library (python implementation of anitomy). I find that they both sometimes miss stuff but together have good coverage. Maybe this would be better as a feature rather than default?

Sorry for any sloppy coding, I don't normally write python code.

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -786,9 +786,11 @@ class Engine:
             # on the user's list. Cache the information.
             # If it fails, cache it as None.
 
-            (show, show_title, show_ep_start, show_ep_end) = utils.guess_aie_anime_information(filename, tracker_list)
+            (show, show_title, show_ep_start, show_ep_end) = utils.guess_anitopy_anime_information(filename,
+                                                                                                   tracker_list)
             if show is None:
-                (show, show_title, show_ep_start, show_ep_end) = utils.guess_anitopy_anime_information(filename, tracker_list)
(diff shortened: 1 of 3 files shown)


z411:
Thanks for your work. Sorry for the late response. I am heavily considering this feature, but as this is done (trying Anitopy *then* AIE) would theoretically double the list scan times at worst. Which is why I'd prefer to choose one or the other; I'll do some testing and if Anitopy does a better job than AIE then I'd prefer to replace it completely. Or at the very least, let the user choose either.

MisterMcDuck:
It's potentially even less performant, because if the path contains the folder name it also tries to run the detection logic on those as well (see https://github.com/z411/trackma/pull/473/commits/d05a8579561a2b428b2e3e47055d69bc03997b7b). I came at this change from a maximum detection perspective, but I can certainly see why you wouldn't want to introduce slowdowns for this process. The folder detection logic did help slightly, but it was minimal and generally only for non-standard release formats. 

It's purely anecdotal, but in my experience Anitopy seemed to perform better with my collection. With that said, there were scenarios where Anitopy failed to detect the release while AIE did.

FichteFoll:
I intended to build a test suite based on some examples from #413 when tackling the recognition issues I had, but I never got around to it. If anitopy as the more general approach could be enhanced with that, I belive it would make sense to try to find problems with it currently and fix them there.

Unfortunately, anitopy and anitomy need to be synchronized manually due to them being implemented in different languages, but the hope would be that both implementations would benefit from that.

PR #476, opened 2020-02-24
CLI: Show status in info instead of numeric value

BamBalaam:
When requesting the info of a tracked item, the CLI returns the internal value instead of a human-readable string.

Before

After


z411:
Thank you!

PR #480, opened 2020-03-27
Instead of using utils.TYPE_* as int, get type from api directly as enum.

ahmubashshir:


FichteFoll:
What really should happen is turning the ints into str enums.

ahmubashshir:
> What really should happen is turning the ints into str enums, imo.

I'm doing this asap.

ahmubashshir:
@FichteFoll, done implementing the enums. Please review.

ahmubashshir:
Would it be good to use enums instead of constants in utils.py?

trackma/ui/qt/models.py
```diff
@@ -346,14 +342,14 @@ def data(self, index, role):
...
+            if t == 'TV':
                 return QtGui.QColor(202, 253, 150)
-            elif t == utils.TYPE_MOVIE:
+            elif t == 'Movie':
                 return QtGui.QColor(150, 202, 253)
-            elif t == utils.TYPE_OVA:
+            elif t == 'OVA':
                 return QtGui.QColor(253, 253, 150)
-            elif t == utils.TYPE_SP:
+            elif t == 'Special':
```
FichteFoll, line 352: I'm still unhappy with this change and it was als the primary reason for my earlier comment.

To be precise, there should be an enum for these 4 literals, likely inside `utils`. Then the site integrations need to provide mappings to the globally known types, so that the interfaces can have a single source of knowledge they can rely on. Incidentally, this is closer to what it was before this PR.

I while explaining this in more detail, I quickly figured out that reimplementing it how I see it should be would be much quicker. In the process I also replaced all the other constants with enums.

Will open a PR soon.

FichteFoll:
Se #491.

PR #481, opened 2020-03-27
Fix some major and minor issues in Gtk UI.

ahmubashshir:
- Stop spinner on search failure.
- Update Headerbar subtitle and Results view to reflect search failure.
- Reenable UI after finishing 'Scan files' action.

PR #482, opened 2020-03-31
Gtk: Fix about window not closing

### LINKED ISSUE #477: Close button on about menu won't work in GTK
Just open About on option or in context menu while Trackma minimized to system tray, then try to tap/click `close`, the UI isn't closed as it should be.

System information:
Version: Trackma-GTK 0.8.2
DE: XFCE 4.14
WM: XFWM4

lastweakness:
This bug only occured in desktops other than GNOME. Fixes #477 .


lastweakness:
#481 includes a simpler fix and also some more fixes for the GTK GUI.

PR #491, opened 2020-05-30
Replace constants in utils.py with enums

FichteFoll:
I replaced all constants defined in utils.py with enums and also replaced various mappings to either translate to the constants or to a displayable string as well. Everything that needs to be known about these values is now contained within their enum class.

Since the lists are persisted through pickle, I'm not exactly sure how well this transition performs, as I encounter an exception when I accidentally ran my system trackma instead of the one I was developing with, but I did not have any issues when upgrading to the new values. I haven't looked into the persisting code, so please consider this when reviewing.

Supersedes and closes #480.


FichteFoll:
For the record, I've been using this branch ever since I worked on it because the new data format makes it incompatible with unpickling on `master`. I don't know if the direction with this is correct, so since apparently my AniList token has expired, I'll be switching back to `master` as I reenter my account.

FichteFoll:
Due to various changes in other places, this PR is conflicting in a lot of files currently. I can merge/rebase the PR, but only if there is interest in adopting it. Just lmk.

z411:
Yes I think it's a good idea; the enum class wasn't part of the standard library back when I built utils.py so it's a welcome addition. Please let me know if you manage to fix it and I'll merge.

ahmubashshir:
I tried it today, but without custom base class for enums, qt frontend crashes when we search anything(qt frontend compares two enums, which was initially unimplemented).
https://github.com/z411/trackma/pull/480#discussion_r432865380 https://github.com/z411/trackma/pull/480#discussion_r432863210

I implemented custom enum base class in my pr so that we can compare two enums.

z411:
> Since the lists are persisted through pickle, I'm not exactly sure how well this transition performs, as I encounter an exception when I accidentally ran my system trackma instead of the one I was developing with, but I did not have any issues when upgrading to the new values. I haven't looked into the persisting code, so please consider this when reviewing.

Regarding this, any issue can be easily avoided by increasing the `version` number of the `Data` class (https://github.com/z411/trackma/blob/e02c82cb60b99751b40788a4a862ffe6ce613d9c/trackma/data.py#L39). This will basically force a re-sync and re-download of the list whenever the user starts Trackma again, and should be done whenever there's a major change to the data model like this one in order to avoid the issues you mention.

ahmubashshir:
I think it would have been better if each api provided its own `Type` enum, because each has its own mapping, and `Type` enum is used to sort media and represent type to user. So if we use a generic types mapping, data might be represented to user wrongly.

FichteFoll:
> I think it would have been better if each api provided its own `Type` enum, because each has its own mapping, and `Type` enum is used to sort media and represent type to user.

I honestly wasn't sure how exactly the Type enum values were used, or I simply forgot by now, but assuming this is primarily for the frontends, it makes sense to have a translation layer between the list backends and the frontends so that each of them only needs to implement one translation instead of a matri for each combination. As such, I believe it is better for the shared enum in `utils` to contain all possible types (or at least all types we want to preserve) rather than each provider defining their own.

FichteFoll:
Merged master into my PR. Hopefully I didn't mess things up. I don't have accounts on most of these platforms to test.

Main difference between this and #527 is that I removed most of the duplicated api lookup tables, probably.

ahmubashshir:
Actually, any of the frontends doesn't translate the Type enums at all. 
They're shown using the __repr__ and as for qt frontend, sorted using 
__gt__, __lt__, __le__, __ge__, __eq__ functions. So IMO, utils module 
should provide the base class for Types enum and each backend should 
implement its own mapping.

PR #493, opened 2020-05-30
Unify ui module launching code

FichteFoll:
By introducing a `__main__` submodule in the gtk and qt packages, the
respective UIs can be launched through `python -m trackma.ui.gtk`, just
like has already been possible with cli and curses.

The gtk main function is moved to the __init__ module just like for the
others.

qt4 loader file has been renamed to remove duplication in the module
path ('trackma.ui.qt.qt4ui').

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,9 +39,9 @@ setup(
             'trackma-curses = trackma.ui.curses:main [curses]',
         ],
         'gui_scripts': [
-            'trackma-gtk = trackma.ui.gtk.main:main [GTK]',
+            'trackma-gtk = trackma.ui.gtk:main [GTK]',
             'trackma-qt = trackma.ui.qt:main [Qt]',
-            'trackma-qt4 = trackma.ui.qt.qt4ui:main [Qt]',
+            'trackma-qt4 = trackma.ui.qt.qt4:main [Qt4]',
         ]
     },
     classifiers=[
--- a/trackma/ui/gtk/__main__.py
+++ b/trackma/ui/gtk/__main__.py
@@ -14,19 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-import os
 
-gtk_dir = os.path.dirname(__file__)
-
-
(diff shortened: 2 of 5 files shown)


FichteFoll:
I should probably clarify on the motivation behind this:

I have trackma installed on my system using a package manager and thus running `python trackma/ui/gtk/main.py` usually falls back to the trackma modules on my system instead of the working directory. I didn't think it would do that, but it appears that's how it is. Using `python -m` makes Python properly prioritize the package in the workingdir and allows me to test the code independently of my system version.

I presume you aren't running into this because you didn't install trackma, but a merge of this would be much appreciated.

FichteFoll:
Rebased on current `master`.

FichteFoll:
Thanks, this should make quick testing of dev code easier (for me, anyway).

z411:
After merging this the Qt interface doesn't start for me. Is it an issue on my end?

FichteFoll:
Pretty sure I tested that back then. You mean Qt, not Qt4, right? What's the error?

z411:
> Pretty sure I tested that back then. You mean Qt, not Qt4, right? What's the error?

Yes, Qt. I didn't attempt to find the reason but it just showed Trackma-qt in console and then hangs (not even ^C closed it). If it doesn't happen to you I'll check if some config on my end that's causing it.

v-fox:
> > Pretty sure I tested that back then. You mean Qt, not Qt4, right? What's the error?
> 
> Yes, Qt. I didn't attempt to find the reason but it just showed Trackma-qt in console and then hangs (not even ^C closed it). If it doesn't happen to you I'll check if some config on my end that's causing it.

@z411 Then why did you merge broken patchset and not fix that came the next day ? Current master is unusable for more than a month because of it.

FichteFoll:
I went ahead and merged #608 since I can do that now and it's a simple but effective change. We haven't discussed any PR merge or review procedures since I've been added as a collaborator, so the state of the project is still a bit unclear.

Issue #177, opened 2015-12-01
An "All" tab for showing all the anime in the lsit


Logmytech:
I'd like to have a tab for all the anime in the my list.

z411:
Implemented in Qt interface.

PR #498, opened 2020-06-20
Use sorted and  filtered global ListStore and implement #177 in Gtk UI.

ahmubashshir:
* Created ShowList class which is a child class of Gtk.TreeModelSort.
 * Used ShowList to wrap Gtk.TreeModelFilter which wraps Gtk.ListStore.
 * Implemented required methods in ShowList.
 * Implemented #177.

### DIFF
--- a/trackma/ui/gtk/mainview.py
+++ b/trackma/ui/gtk/mainview.py
@@ -554,7 +554,7 @@ class NotebookPage(Gtk.ScrolledWindow):
 
     @decimals.setter
     def decimals(self, decimals):
-        self._show_tree_view.decimals = decimals
+        self._list.decimals = decimals
 
     @property
     def status(self):
--- a/trackma/ui/gtk/mainview.py
+++ b/trackma/ui/gtk/mainview.py
@@ -99,6 +99,9 @@ class MainView(Gtk.Box):
         self.image_container_box.pack_start(self.image_box, False, False, 0)
 
         self.notebook.set_scrollable(True)
+        self.notebook.add_events(Gdk.EventMask.SCROLL_MASK |\
+            Gdk.EventMask.SMOOTH_SCROLL_MASK)
+        self.notebook.connect('scroll-event', self._notebook_handle_scroll)
 
         self.statusbar = Gtk.Statusbar()
         self.statusbar.push(0, 'Trackma-gtk ' + utils.VERSION)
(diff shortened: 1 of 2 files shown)


ahmubashshir:
@FichteFoll,@z411 can you review these changes, please?

FichteFoll:
I don't have any expertise with this part of the code, nor with Gtk, so I'll refrain from reviewing. I did take a brief look, though.

ahmubashshir:
> I don't have any expertise with this part of the code, nor with Gtk, so I'll refrain from reviewing. I did take a brief look, though.

Then can you test and notify me if you find any bugs, please?

z411:
Great, a global filtered ListStore is the way to go for this and it's something I always wanted to do and it reaches feature parity with the Qt interface. I plan to go back to development soon so I'll take a look as soon as I can. Thanks for your work.

fushinari:
I tested it and it works perfectly for me.

ahmubashshir:
> I tested it and it works perfectly for me.
> 
> Edit: Can you make the number of entries optional?

Ok.

fushinari:
> Ok. btw, I tried to follow the qt ui.

I like that it's there. I just don't find myself needing it. I feel like that information should be presented elsewhere but I don't know where either.

ahmubashshir:
@fushinari should I add a field 8n ui-gtk.json for this?

fushinari:
I personally would've liked that but I'm not sure how other would feel. Right now, tab scrolling is disabled, so the window size becomes fixed (at around 920px for me with the counters and even without that, it was fixed at around 620 or so). Even with the tab scrolling enabled, it will become fixed at around 570px. And that would've been great if Notebook wasn't wasting space just because scrolling is enabled.

I tried enabling GtkNotebook's tab scrolling:

Screenshot from 2020-10-21 01-05-09

I don't know if it's better or worse tbh... I'll leave that up to you ig. I'm fine with that myself though.

z411:
The window indeed gets too wide, so I think tab scrolling is a good idea.

ahmubashshir:
@z411 @fushinari enabled scrolling in GtkNotebook(with mouse scroll support).

ahmubashshir:
BTW, @fushinari anime is shown twice in ui. Shouldn't it appear in only subtitle?

image

fushinari:
Oh, sorry I missed that. But yeah it's redundant, you could remove it.

z411:
Thank you.

FichteFoll:
The individual tabs seem to become unnecessarily large when the window is too small, but that's padding is only added when not all tabs can fit into the window and made so that the tabs that didn't make it aren't cut off, so this is fine to me. The most important tabs are at the front anyway.

Issue #214, opened 2016-06-18
Tracker support for Kodi


lolihunter:
I have an android TV box that I use for pretty much all my anime and tv. I have a server running Arch, and have tried to get the tracker to be able to detect when I am watching the shows to no avail. Trying to set the tracker to pick up on samba reading the file doesn't work, (smbd process).

z411:
So are you running Trackma on the server, correct?
Implementing Kodi support shouldn't be hard as we already have a Plex implementation thanks to a contributor (see extras/plex.py).

lolihunter:
Yes, running trackma on an Arch linux server, and using Kodi on android. Kodi has no way to scrobble to hummingbird itself.

lolihunter:
Oh, and FYI, kodi does support control through a webserver (if enabled) on port 8080. Although it might take some work to implement a way for trackma to detect what is playing and make it work.

z411:
Implemented by #214 - Thank you.

PR #509, opened 2020-09-04
Kodi tracker support

Hairo:
Kodi tracking support, only works with the internal player (for now) and requires enabling `Settings -> Services -> Control -> Allow control of Kodi via HTTP` to work. 

Only tested with Kodi 18.8 but should work with any version that uses the [JSON-RPC API/v10](https://kodi.wiki/view/JSON-RPC_API/v10).

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -986,6 +986,9 @@ class Engine:
         if ttype == 'plex':
             from trackma.tracker.plex import PlexTracker
             return PlexTracker
+        elif ttype == 'kodi':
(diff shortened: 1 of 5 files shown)


Hairo:
Can close #214 with this.

Hairo:
Using the kodi 19 testbuilds it seems to work with the [JSON-RPC_API/v12](https://kodi.wiki/view/JSON-RPC_API/v12) too.

z411:
I don't have the means to test it but so far it seems it doesn't break any existing functionality so I'll merge. Thanks a lot for the work.

PR #510, opened 2020-09-13
Fix some issues with the GTK GUI

fushinari:
Made several changes that are not immediately visible. The commit messages are hopefully descriptive enough.

- Now uses a GtkPopover for setting episode number.
- Use a GtkButtonBox for the episode number, "+" and "-" buttons.
- Stop using "Gtk.STOCK_" items because they've been deprecated and use suggested alternatives from GTK docs.
- Library info (i.e. "Anilist [manga]") is now in the subtitle instead of the title and is now updated properly.
- Some spacings have been fixed to match the GNOME HIG and results in a more pleasant UI.

There are a few changes that are personal opinions and can be reverted if necessary:
- "Trackma-gtk" is less suited for use anywhere other than the CLI. So, I've renamed all instances of it within the UI to "Trackma GTK" and just "Trackma" in the application title. Now, it only mentions "GTK" in the About window and the stdout.
- Stop showing the version info in title bar because it's available in the About window anyway.

The UI looks like this now:  
Screenshot from 2020-09-21 00-07-23

### DIFF
--- a/trackma/ui/gtk/accountswindow.py
+++ b/trackma/ui/gtk/accountswindow.py
@@ -17,8 +17,7 @@
 import os
 import webbrowser
 from enum import Enum
-from gi import require_version
-require_version('Gtk', '3.0')
+
 from gi.repository import Gtk, GdkPixbuf, GObject
 from trackma.ui.gtk import gtk_dir
 from trackma.ui.gtk.gi_composites import GtkTemplate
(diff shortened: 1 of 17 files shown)


fushinari:
I also think that the screenshot needs to be updated, especially since Adwaita has changed so much since the current screenshot.

fushinari:
Would it be better to remove the "GTK" entirely from the application title? I feel like it's unnecessary.

Wouldn't it also be better to show the Trackma icon on startup in the imagebox instead of the "Trackma" text?

fushinari:
Screenshot from 2020-09-20 23-45-27

This is how it now looks on launch.

ahmubashshir:
> Screenshot from 2020-09-20 23-45-27
> 
> This is how it now looks on launch.

Please use default font and theme while submitting ss.

fushinari:
> Please use default font and theme while submitting ss.

That wasn't meant an ss submission, just as a demo.

| On Launch | With Selected Item | With Menu Open
| ------------ | ------------------ | --------------- |
| Screenshot from 2020-09-21 00-07-12 | Screenshot from 2020-09-21 00-07-20 | Screenshot from 2020-09-21 00-07-23 |

fushinari:
Is there a symbolic version of the icon?

z411:
Looks nice. Thanks a lot for your work.

PR #516, opened 2020-11-03
[Plex Tracker] Update state on plex video close

slapelachie:
Found an issue while using the plex tracker where upon closing the video the tracker would not update, and would continue to think the video was playing. I implemented a quick and dirty fix to allow for the update to occur. I also made hook calls print in debug mode as tracker_state spams the screen every tracker_interval


z411:
Thank you.

PR #519, opened 2020-11-12
Added Jellyfin Tracker Support

slapelachie:
Wanted trackma to keep track of my Jellyfin instance so I decided to fully implement it into the source code. I also added a view offset setting for the session info, I implemented this for one of my discord hooks to display where I am in the show. 

Some things to note for documentation:
- The API key in the config file is found in Jellyfin under Dashboard -> API Keys, where the user needs to generate this key and plug it into the config file. It seems that this can only be generated under an admin user
- The username in the config file is what user the user wants to track


z411:
Looks good, thank you. Also a Discord hook sounds interesting; do you have it working somewhere?

slapelachie:
The library I am using ([pypresence](https://pypi.org/project/pypresence/)) crashes the GTK frontend for some reason (will just continuously try to import the hook), so it's pretty unstable. I've got it hosted [here](https://gist.github.com/slapelachie/5d6d8b912632625caff07016355996f2)

PR #521, opened 2020-11-12
[Gtk UI] Tooltips

ahmubashshir:
* Add functions for properly showing tooltip
* Change Class structure for easier data retrieval


z411:
Thank you for your work on the GTK interface!

Issue #525, opened 2020-11-15
Playing episodes stopped working


FichteFoll:
Some time ago (unfortunately I don't remember when exactly), playing a file stopped working without me modifying the config. I suspect it was an update, but I want to confirm the issue first.

Using `trackma -d`, the following is printed on the terminal, but nothing happens.

```
FichteFoll [anilist] (anime) watching >> play 2
Engine: Getting Adachi and Shimamura 4 from library...
Engine: Found. Starting player...
FichteFoll [anilist] (anime) watching >>
```

It also does not work with the gtk frontend. I'm using a build fresh off `master` and my player config is `"mpv"`.

I will bisect later if you can't reproduce.

FichteFoll:
I forgot to mention I was on Arch Linux.

Anyway, browsing my notifications, I spotted https://github.com/z411/trackma/pull/513, which may be related to my problem. I'll verify that later when I find the time.

z411:
Can reproduce. As an aside, the Qt interface hard locks up when trying to play an episode. Seems severe. Will look into this as well.

ahmubashshir:
Gtk interface works, command line works too... launched player isn't being forked off in qt interface, curses interface doesn't even starting the player process... I'm using my [local][1] branch.

[1]: https://github.com/ahmubashshir/trackma/tree/local

ahmubashshir:
After searching about `QThread` and `os.fork` I found out that they doesn't work well together. I'm working on it

z411:
To fix this we need help in the #508 to reproduce a bug. Anyone willing to test the `prevent-zombie` branch to see if closing Trackma also closes any player started by Trackma is welcome.

FichteFoll:
I just tested `prevent-zombie` and `play` works and the player is closed when I ctrl-c in trackma (i.e. close it).

FichteFoll:
I was able to determine the cause by my player setting being set to simply `mpv` where the spawn code uses `os.execv`, which *does not perform a PATH lookup*. Changing my setting to `/usr/bin/mpv` makes the player process behave as expected in that it outlives trackma when it is closed but also does not come up in `zps -l` when killed before mpv. Monitoring the process tree with `htop` also reveals that the player process is parented to the init process.

I suggest to 

1. (optionally) wrap the player configuration with `shutil.where`
2. check if the specified path exists and fail early
[...]

PR #526, opened 2020-11-17
[Engine+Qt] Fixes

ahmubashshir:
Engine:
 * support external function for creating detached process
 
Qt:
 * Launch player using `QtCore.QProcess`

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -74,10 +74,11 @@ class Engine:
                'tracker_state':     None,
                }
 
-    def __init__(self, account=None, message_handler=None, accountnum=None):
+    def __init__(self, account=None, message_handler=None, accountnum=None, start_process=None):
         self.msg = messenger.Messenger(message_handler)
 
         # Utility parameter to get the account from the account manager
+        self._start_process = start_process
         if accountnum:
             from trackma import accounts
             account = accounts.AccountManager().get_account(accountnum)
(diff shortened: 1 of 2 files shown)


ahmubashshir:
I think this will fix #525

z411:
This issue made me think about this but, instead of passing a custom function to the engine, maybe it would be better for the user interface to handle starting processes. The engine can just return the process and the arguments (or just the filename?) and the actual starting could be handled by the user interface. This is already the case for opening links for example.

ahmubashshir:
> This issue made me think about this but, instead of passing a custom function to the engine, maybe it would be better for the user interface to handle starting processes. The engine can just return the process and the arguments (or just the filename?) and the actual starting could be handled by the user interface. This is already the case for opening URLs for example.

We can define a signal handler for launching player with a default handler. Then engine will call connected signal handler with `player` and `filename` args.

z411:
Thanks for your work. I modified this so the spawning of subprocesses is handled by the UIs as I said. Sadly, the problem still isn't fixed. Does it work at all for you? It works for me on Qt only, but the double fork method (I moved it to `utils.py`) does nothing for me. I pushed to the `ahmubashshir-prevent-zombie` branch.

z411:
Thanks for your work. I modified this so the spawning of subprocesses is handled by the UIs as I said.

Sadly, the problem still isn't fixed. Does it work at all for you? It works for me on Qt only, but the double fork method (I moved it to utils.py) does nothing.

ahmubashshir:
Didn't tried yet. I'll notify when I try it :)

PR #529, opened 2020-11-26
use pydbus instead of dbus-python in MPRIS Tracker

ahmubashshir:
As `dbus-python` is deprecated, I replaced it with pydbus in MPRIS tracker.


z411:
Thank you!

PR #542, opened 2021-01-11
[HOOK] pypresence

ahmubashshir:
- Added hook for discord rich presence using pypresence.
- Added assets for discord rich presence hook.


z411:
Nice work. Thank you.

PR #544, opened 2021-01-14
[HOOK] DiscordRPC update

ahmubashshir:
* Use thread to retry connection in case of failure.


hooks/presence.py
```diff
@@ -0,0 +1,152 @@
...
+        }
+        self._update = True
+
+    def run(self):
+        set_loop(new_loop())
+        while True:
+            try:
+                self._reconnect()
+                if self._enabled:
+                    if self._update:
```
FichteFoll, line 77: You can combine these conditions,.

hooks/presence.py
```diff
@@ -0,0 +1,152 @@
...
+        episode = status["show"][-1]
+        total = show["total"] or estimate_aired_episodes(
+            engine.get_show_info(show['id'])
+        ) or '?'
+
+        if status["paused"]:
+            rpc.present(engine,
+                        status["viewOffset"],
+                        "Paused {}".format(title),
+                        "Episode {} of {}".format(
```
FichteFoll, line 136: I've noticed this while rebasing my other PR earlier, but does this repo use/enforce 80 character max width and do we think this is contemporary? I personally prefer going up to 120 instead of wrapping eagerly like this.

ahmubashshir:
dunno... personally, I try to follow the 80 columns limit.

hooks/presence.py
ahmubashshir: I prefer to follow pep-8 style guideline.

hooks/presence.py
ahmubashshir: combined.

Issue #532, opened 2020-12-06
Error when searching for Nichijou on MAL


Skaytacium:
When searching for "Nichijou" or "nichijou" or "nichi" it raises a KeyError: 'unknown'

` [mal] (anime) watching >> add Nichijou
libmal: Searching for Nichijou...
Traceback (most recent call last):
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\Scripts\trackma-script.py", line 33, in <module>
    sys.exit(load_entry_point('Trackma==0.8.4', 'console_scripts', 'trackma')())
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\site-packages\trackma-0.8.4-py3.9.egg\trackma\ui\cli.py", line 1079, in main
    main_cmd.cmdloop()
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\cmd.py", line 138, in cmdloop
    stop = self.onecmd(line)
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\site-packages\trackma-0.8.4-py3.9.egg\trackma\ui\cli.py", line 780, in onecmd
    return self.execute(cmd, args, line)
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\site-packages\trackma-0.8.4-py3.9.egg\trackma\ui\cli.py", line 797, in execute
    return func(args)
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\site-packages\trackma-0.8.4-py3.9.egg\trackma\ui\cli.py", line 409, in do_add
    entries = self.engine.search(args[0])
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\site-packages\trackma-0.8.4-py3.9.egg\trackma\engine.py", line 482, in search
    return self.data_handler.search(criteria, method)
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\site-packages\trackma-0.8.4-py3.9.egg\trackma\data.py", line 219, in search
    results = self.api.search(criteria, method)
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\site-packages\trackma-0.8.4-py3.9.egg\trackma\lib\libmal.py", line 318, in search
    results.append(self._parse_info(item['node']))
[...]

PR #545, opened 2021-01-20
Some titles may return 'unknown' media_type

mobrine1:
Related to #532


z411:
Thank you!

FichteFoll:
Shouldn't 'unknown' map to … the unknown type? (conflicts with #491)

mobrine1:
Yeah, if you say so.
I used `other`, because `media_type` is defined.
The only example I did found where `utils.TYPE_UNKNOWN` is where `None` is returned.

PR #547, opened 2021-02-07
Tweak show recognition code and add tests

FichteFoll:
By adding some test cases I collected on a certain website and from #413, I started rewriting a couple matchings in the show parsing algorithm without touching too much of the overarching logic, because that seemed to have worked quite well outside of the collected cases.

Tests are run using `pytest` in the repository root. I didn't add anything for documentation, however, since this is the only part that has tests now.

I tried to make the changes modular and have each commit stand on its own. Particularly notable changes are:

1. S01E01 parsing, where the season number is extracted into a separate attribute and then simply appended to the final show name because I didn't know what else to do it without refactoring. See https://github.com/z411/trackma/commit/c25077c19ff9644ff6cfb30c42dfb7b0161707df for more details. Works decently well for me, anyway, and where it doesn't, altnames help. For matching [Tensei shitara Slime Datta Ken S02E04](https://anilist.co/anime/108511/That-Time-I-Got-Reincarnated-as-a-Slime-Season-2/) properly without an altname, further steps are needed.
2. Not stripping years in parentheses.
3. Tweaking the episode number finder so it only catchs truly isolated episode numbers (i.e. not within an episode title).

Files from #413 I didn't handle because they aren't proper episodes:
```
[Opportunity] The Tatami Galaxy - NCED [BD 720p] [903A3AD6].mkv
[-__-'] Girls und Panzer NCED 04 [BD 1080p FLAC] [88F40C98].mkv
[-__-'] Girls und Panzer NCED OVA Anzio-sen [BD 1080p FLAC] [ED2078EC].mkv
[-__-'] Girls und Panzer NCOP [BD 1080p FLAC] [04ED15D5].mkv
```
We should discuss how to proceed with these (here or later).

Shouldn't conflict with #473, but I would not merge that as-is regardless since it has a lot of inefficient code duplication.

### DIFF
--- a/tests/test_anime_info_extractor.py
+++ b/tests/test_anime_info_extractor.py
@@ -88,7 +88,7 @@ def test_dubsub():
         extension="mkv",
         resolution="1080p",
         audioType=["AAC"],
-        releaseSource=["WEB"],
+        releaseSource=["Web-DL"],
     )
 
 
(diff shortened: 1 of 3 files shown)


FichteFoll:
bump

z411:
Seems like a lot of work. Thank you. Indeed this might end up being deprecated by Anitopy but as I suggested in #473 we'll probably try having both at the same time for a while to see which one works and performs better. Did you see any drop in performance after these changes?

PR #554, opened 2021-03-14
[Gtk] AccountDialog headerbar fix.

### LINKED ISSUE #553: First launch is broken in GTK GUI
The AccountsWindow object does not properly get initialized and so, the user becomes unable to add an account. Seems related to #538 .

image

ahmubashshir:
- [GTK] Force headerbar in AccountDialog, closes #553


fushinari:
Weird... but it does fix the issue...

ahmubashshir:
ping @fushinari try the latest commit, it fixed the info window.

fushinari:
Yes, everything seems to work as expected now! Thanks for the quick fix!

PR #586, opened 2021-10-25
[Tracker/polling] Return full path if `library_full_path` is enabled

mnabid:
Allow **`polling (lsof)`** tracker to scan the full path when `library_full_path` is enabled in config.

**`inotify`** tracker [already does this](https://github.com/z411/trackma/blob/55728fde0cd3c58044e9f2b7d6d1e56fe189148b/trackma/tracker/inotifyBase.py#L122-L125).


FichteFoll (approved):

mnabid:
Closing in favor of #592.

PR #589, opened 2021-11-27
Extract episode numbers from notation like S02E05

Self-Perfection:
Such notation seems to be widespread.

Also I'd like to note that this massive regex logic is hard to understand and debug. It really looks like this AnimeInfoExtractor should have unit tests.

### DIFF
--- a/trackma/extras/AnimeInfoExtractor.py
+++ b/trackma/extras/AnimeInfoExtractor.py
@@ -258,9 +258,9 @@ class AnimeInfoExtractor():
         if not self.episodeStart:
             # Check if there is an episode specifier
             m = re.search(
-                '(?:[^0-9a-zA-Z])(E\.?|Ep(?:i|isode)?(?: |\.)?)(\d{1,}(?:\.\d{1})?)(?:[^\d]|$)', filename, flags=re.IGNORECASE)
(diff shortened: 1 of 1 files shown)


trackma/extras/AnimeInfoExtractor.py
```diff
@@ -258,9 +258,9 @@ def __extractEpisodeNumbers(self, filename):
         if not self.episodeStart:
             # Check if there is an episode specifier
             m = re.search(
-                '(?:[^0-9a-zA-Z])(E\.?|Ep(?:i|isode)?(?: |\.)?)(\d{1,}(?:\.\d{1})?)(?:[^\d]|$)', filename, flags=re.IGNORECASE)
+                '(?:[^0-9a-zA-Z])(S[0-9]+)?(E\.?|Ep(?:i|isode)?(?: |\.)?)(\d{1,}(?:\.\d{1})?)(?:[^\d]|$)', filename, flags=re.IGNORECASE)
```
ahmubashshir, line 261: 👍🏽

ahmubashshir:
> Such notation seems to be widespread.
> 
> Also I'd like to note that this massive regex logic is hard to understand and debug. It really looks like this AnimeInfoExtractor should have unit tests.

Not that massive though :3

Self-Perfection:
Whoa, I've noticed there is big #547  with refactoring and test. It also includes  S01E01 parsing.
Well apparently this PR becomes obsolete if #547 get merged.

z411:
#547 got merged. Thank you!

PR #591, opened 2021-12-31
Replace AnimeInfoExtractor with Anitopy

mnabid:
This PR replaces Trackma's own anime info parser with [Anitopy](https://github.com/igorcmoura/anitopy), a python implementation of Taiga's [Anitomy](https://github.com/erengy/anitomy).

Unlike #473, I added Anitopy as a drop-in replacement for AnimeInfoExtractor. AIE has a lot of parsing issues and the code is hard to debug with all the complicated regexps. #547 and #589 fixed some issues for me, but overall Anitopy works much better with my library.

I also dealt with some stuff that Anitopy doesn't―
- **`Season X Part Y` detection**
Anitopy recognizes `Part Y` as episode title.

- **Fixing anime types (`OVA`, `Special` etc.) being detected as episode title**
Might happen where `Season X Part Y` is involved.

- **Ignoring openings, endings, promotionals and other "non-episodes"**
They usually come as bonus materials with DVD, and don't need to be tracked. Also, MAL and Anilist almost always ignore them.

- **Parsing non-standard episode numbers such as `S01OVA05`, `S01S05` etc.**
Judas uses this format a lot (*One-Punch Man*, *Mob Psycho 100*, *Shirobako*).
Anitopy can parse `S01E05` though, but not the aforementioned ones.

- **Dealing with filenames that has both consecutive and seasonal episode numbers**
Anitopy can detect both of them, so I only kept the consecutive episode number.
This would be useful if we're dealing with a long-running weekly anime, e.g. `[Judas] Black Clover - S02E01 (052).mkv`
Even if the anime is seasonal, chances are `anime-relations.txt` has a rule for it.

- **Folder detection**
Parsing folder name and filename separately and choosing one over the other is probably not a good idea, because which one to choose might be a problem. [#473](https://github.com/z411/trackma/pull/473) implemented this by running `guess_show()` on each of them, which is sloooooowww.
What I did was― if folder name and filename share a common part, I removed the duplicate part and then melded the remaining parts together.
Probably not a good idea either (and I did it in a half-assed way), but works for me so far :shrug:

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -790,7 +790,7 @@ class Engine:
             for fullpath, filename in utils.regex_find_videos(searchdir):
                 if self.config['library_full_path']:
                     filename = self._get_show_name_from_full_path(
-                        searchdir, fullpath).strip()
+                        searchdir, fullpath)
                 (library, library_cache) = self._add_show_to_library(
                     library, library_cache, rescan, fullpath, filename, tracker_list)
 
@@ -1024,8 +1024,8 @@ class Engine:
 
     def _get_show_name_from_full_path(self, searchdir, fullpath):
         """Joins the directory name with the file name to return the show name."""
-        relative = fullpath[len(searchdir):]
-        return relative.replace(os.path.sep, " ")
+        relative = fullpath[len(searchdir):].lstrip(os.path.sep)
+        return relative
 
     def _searchdir_exists(self, path):
         """Variation of dir_exists that warns the user if the path doesn't exist."""
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -438,7 +438,7 @@ class Engine:
 
             aie = AnimeInfoExtractor(filename)
             (show_title, ep) = aie.getName(), aie.getEpisode()
-            self.msg.debug(self.name, "Guessed {}".format(show_title))
+            self.msg.debug(self.name, "Show guess: {}".format(show_title))
 
             if show_title:
                 tracker_list = self._get_tracker_list()
@@ -856,6 +856,7 @@ class Engine:
                 if show:
                     self.msg.debug(
                         self.name, "Adding to library: {}".format(fullpath))
+                    self.msg.debug(self.name, "Show guess: {}".format(show_title))
 
                     if show_ep_start == show_ep_end:
                         # TODO : Support redirections for episode ranges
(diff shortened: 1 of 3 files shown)


mnabid:
CC @FichteFoll, @MisterMcDuck, @ahmubashshir

z411:
Good stuff. My suggestion though, instead of working on the AnimeInfoExtractor, is to make a separate class, let's call it `AnitopyWrapper`. In that case for now we can have both implemented for a while and let the user choose either, and whichever ends up giving best results for everyone, we deprecate the worst one.

In other words, could you move the changes to `AnitopyWrapper.py` instead and leave the original version of `AnimeInfoExtractor.py` there as well? After that I'll merge and work on a wrapper to choose one of the two depending on user choice/library availability.

mnabid:
@z411 Done.

z411:
This is also breaking the newly merged enums in tracker.py - could you attempt to merge from master?

imsamuka:
@mnabid, the configuration `library_full_path` creates a filename that can't be parsed by AnitopyWrapper. When enabled, the filename includes its relative path like `'animeFolder/[Fansub] Anime Name.mkv'`). It was driving me nuts, i'm so glad to find it.

I'm studying the AIE code to find the best way to counteract this issue.

FichteFoll:
I finally found the time to test this a bit. I also applied some changes to allow switching between aie and anitopy via config. You can find the changes on the [`feature/anitopy` branch](https://github.com/z411/trackma/tree/52ade32e8bb46ed3b78875bee2c4cbbbbcee595f).

- I didn't notice much of a change compared to aie, neither from functionality nor from speed., I suppose adding the season of a show as "Season N" instead of simply "N" yields better results for most cases. Not really related to anitopy, though.
- "anime type" matching looks useful for detecting NCOP and similar files that don't represent episodes. This works.
- Anitopy seeminly cannot detect episodes like `07v1.5`, whereas `06v2` is properly redirected to 6.
- Couldn't match `Summer Time Render (Summertime Rendering) [JapDub,EngSub,SlyFox]/[SlyFox] Summertime Rendering - 01 [33073356].mkv`
- Primarily using dynamic access to an object's attributes or even `__dict__` is an anti-pattern. It would be easier to simply keep and operate on anitopy's result in a proper dict.
- The tests we have for aie should be adapted for anitopy, especially to catch potential regressions.

PR #596, opened 2022-01-16
[Kitsu API] Deals with the case where there is no small/tiny image

Kcchouette:
In the case of a recent anime, here the content we have back for poster:
```
      "posterImage": {
        "original": "https://kitsu-production-media.s3.us-west-002.backblazeb2.com/anime/44979/poster_image/9f6b8e791778abde38ca7e711327f50e.jpg?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=002d6013655cdc00000000003%2F20220116%2Fus-west-002%2Fs3%2Faws4_request&X-Amz-Date=20220116T214041Z&X-Amz-Expires=900&X-Amz-SignedHeaders=host&X-Amz-Signature=602c57bd3964a1aa4074cefb1f0e39d30a1ff6b46d23cd703cc12097b70959df",
        "meta": {
          "dimensions": {}
        }
      }
```
(for CUE anime / id = 44979)

So we have no small/tiny poster and so trackma keeps crashing

### DIFF
--- a/trackma/lib/libkitsu.py
+++ b/trackma/lib/libkitsu.py
@@ -353,7 +353,7 @@ class libkitsu(lib):
                         'my_id': entry['id'],
                         'my_progress': entry['attributes']['progress'],
                         'my_score': float(rating)/4.00 if rating is not None else 0.0,
-                        'my_status': entry['attributes']['status'],
+                        'my_status': status,
                         'my_start_date': self._iso2date(entry['attributes']['startedAt']),
                         'my_finish_date': self._iso2date(entry['attributes']['finishedAt']),
                     })
@@ -560,8 +560,8 @@ class libkitsu(lib):
             # For now I'm just picking the romaji title in these cases.
             'title':       attr['titles'].get('en_jp') or attr.get('canonicalTitle') or attr['titles'].get('en'),
             'total':       total or 0,
-            'image':       attr['posterImage'] and attr['posterImage']['small'],
-            'image_thumb': attr['posterImage'] and attr['posterImage']['tiny'],
+            'image':       attr['posterImage'] and attr['posterImage'].get('small'),
+            'image_thumb': attr['posterImage'] and attr['posterImage'].get('tiny'),
             'start_date':  self._str2date(attr['startDate']),
             'end_date':    self._str2date(attr['endDate']),
             'type':        self.type_translate.get(attr['subtype'], utils.TYPE_UNKNOWN),


ahmubashshir:
@Kcchouette, try writing more descriptive commit message than just "Update X" or "Fix Y"

Kcchouette:
@ahmubashshir you are right, but when merging you can squash it
Here title is understandable

ahmubashshir:
People do more merge than squash, and the "Update X/Y" commit messages don't sense in that situation.

Fun fact: you didn't have to write a good description and title for the PR at all, if you write commit messages properly, github-cli uses the first commit title as PR title and generates the body from either first commit message body or from the title of all commit unmerged in that branch.

trackma/lib/libkitsu.py
```diff
@@ -353,7 +353,7 @@ def fetch_list(self):
                         'my_id': entry['id'],
                         'my_progress': entry['attributes']['progress'],
                         'my_score': float(rating)/4.00 if rating is not None else 0.0,
-                        'my_status': entry['attributes']['status'],
+                        'my_status': status,
```
ahmubashshir, line 356: Good catch, I forgot to use the `status` variable here when I commited changes.

trackma/lib/libkitsu.py
```diff
@@ -560,8 +560,8 @@ def _parse_info(self, media):
             # For now I'm just picking the romaji title in these cases.
             'title':       attr['titles'].get('en_jp') or attr.get('canonicalTitle') or attr['titles'].get('en'),
             'total':       total or 0,
-            'image':       attr['posterImage'] and attr['posterImage']['small'],
-            'image_thumb': attr['posterImage'] and attr['posterImage']['tiny'],
+            'image':       attr['posterImage'] and attr['posterImage'].get('small'),
```
FichteFoll, line 563: Should we have a fallback to 'original' if 'small' cannot be found?

z411: `image` is used by the details dialog. How small is the small thumb? And yes I think a fallback to original would be sensible as the details dialog can be as big as you want to.

Eiji7:
@FichteFoll Correct me if I'm wrong, but looks like this PR is no longer needed. I have created more detailed issue #665 and it was fixed in PR #666. I'm aware that you have created PR I linked, but maybe you have just forgot about this one (which is nothing surprising having in mind there is 1 year difference)?

FichteFoll:
Yes, thanks for the reminder. I don't keep track of all the old issues on this repo on a regular basis.

Issue #598, opened 2022-01-21
Full Rescan CPU load


serjflint:
Clicking Rescan Library (full) eats one core 100% for several minutes and the QT interface barely responds at the time. I have ~4k titles in total on the lists but barely 100 files on the disk.

serjflint:
![2022-01-21_22-17](https://user-images.githubusercontent.com/4900246/150587105-68d2e08e-6141-43eb-a945-810cf057c41d.png)
Obvious suspect is guess_show with SequenceMatcher

serjflint:
I created a dirty hack to scan only currently watching titles instead of the whole library when scan_whole_list is disabled.

ahmubashshir:
we should scan everything in a separate thread and collect the result after it finishes.

serjflint:
> We should scan everything in a separate thread and collect the result after it finishes.

It solves the symptoms, but not the cause. There are many optimizations to find the match than just brute-force seeking longest common sub-sequence of each file and each title.
1. normalizing the titles (both files and library, calling lower() is just the beginning)
2. hash-table for library (like for alternative titles)
3. limit the candidates (like I've done in PR above, but there can be done more)

FichteFoll:
Indeed, matching is quote slow especially when there is no match for a file.

z411:
> I created a dirty hack to scan only currently watching titles instead of the whole library when scan_whole_list is disabled.

It is not scanning the whole library, just watching, on hold and plan to watch, as you can see here:

https://github.com/z411/trackma/blob/master/trackma/lib/libmal.py#L53

serjflint:
@z411 sorry, I said it incorrectly. I agree, it is not the whole library. But in my case the difference is negligible. I have 90% of titles in "plan_to_watch". For 2k titles the current brute-force algorithm is horribly slow. I mentioned some ideas [above](https://github.com/z411/trackma/issues/598#issuecomment-1019312633).

z411:
Limiting the candidates is what's being done now, but I think the main fix would be to make the algorithm faster. As my list was never too big I never noticed horrifying performance but admittedly Taiga is a lot faster even though it's scanning the whole list; we really need to take from that. Or maybe it's not that faster, it's just that as @ahmubashshir said it's being done in a separate thread.

PR #599, opened 2022-01-22
fix: scan only currently watching when scan_whole_list is disabled

serjflint:
Relates: #598

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -773,7 +773,7 @@ class Engine:
                 my_status = self.mediainfo['statuses']
             else:
                 my_status = self.mediainfo.get(
-                    'statuses_library', self.mediainfo['statuses_start'])
+                    'statuses_start', self.mediainfo['statuses_library'])
 
         if rescan:
             self.msg.info(


FichteFoll (approved):

z411:
This was the original behavior. This was changed because people wanted also to scan on hold and plan to watch lists as well, as they usually don't have a lot of results in them either.

serjflint:
@z411 they can use "Scan through whole list" for that, though.

z411:
> @z411 they can use "Scan through whole list" for that, though.

Scan through whole list also scans through Finished, which is most of the time, a ton. This PR removes a feature that was requested. In that case we could add three options: Scan only watching, scan usual lists (watching, on hold, plan_to_watch), scan everything (edge case, probably).

serjflint:
@z411 I agree. The option to choose what to scan is better. That's why I called this PR 'a dirty hack' in the [discussion](https://github.com/z411/trackma/issues/598#issuecomment-1019214648)  =)

FichteFoll:
> Scan only watching, scan usual lists (watching, on hold, plan_to_watch), scan everything (edge case, probably).

Would making the list of statuses to scan configurable be feasible? That would certainly be the most flexible options and I think also be more transparent for the user since the current behavior even confused me in the code (i.e. what do `statuses_library`, `statuses`, and `statuses_library` mean exactly?).

serjflint:
@FichteFoll better move to the related [issue](https://github.com/z411/trackma/issues/598). I closed this PR and we need a new one to solve the issue the way you mentioned.

Issue #600, opened 2022-02-05
Is this compatible with ani-cli?


Merulox:
Out of the box it doesn't seem to be.. Is it just that I'm doing something wrong?

https://github.com/pystardust/ani-cli

ahmubashshir:
You'll have to use [mpv-mpris](https://github.com/hoyon/mpv-mpris) and mpris tracker.

Merulox:
Sorry, could you elaborate on what mpris tracker is

ahmubashshir:
[MPRIS Arch wiki entry](https://wiki.archlinux.org/title/MPRIS).

[XDG Specification](https://specifications.freedesktop.org/mpris-spec/latest/)

Merulox:
![image](https://user-images.githubusercontent.com/68356319/153734367-6ce19f83-36dd-44a5-9d85-7c0045ae0023.png)

is that good enough for trackma to guess?

FichteFoll:
I doubt it since the mix of spaces and hyphens isn't accounted for in the title matching code.

iamchokerman:
hi there, ani-cli maintainer. in the latest ani-cli release the title has been changed for better compatibility with trackma, an example would look like this in mpv "Akebi Chan No Sailor Fuku Episode 5"
additionally, if for some reasons externals trackers don't work for you (for example i use OSX, which doesn't have any good mpv trackers afaik), i wrote this short script for trackma synchronization, feel free to tweak it to your needs
```#!/bin/sh
if [ -z "$*" ]; then
  read -p "> " query
else
  query=$*
fi

INDEX="$(trackma -a 1 search "$query"|awk '{print $2}'|sed '2q;d')"
EPNUM="$(trackma -a 1 search "$query"|grep -Eo '[0-9]{1,3} / [0-9]{1,3}'|cut -d "/" -f 1)"
ANIME="$(trackma -a 1 info "$INDEX"|head -n 1)"

ani-cli -a $((EPNUM+1)) "$ANIME"

read -r -p "Do you want to set this episode as watched? [Y/n]" response
if [[ $response =~ ^(yes|y| ) ]] || [[ -z $response ]]; then
   trackma -a 1 update "$INDEX" $((EPNUM+1))
fi
```

iamchokerman:
@Merulox

imsamuka:
You also need to `pip install pydbus`.

RaitaroH:
Consider my adl script.
[...]

PR #610, opened 2022-04-25
Add ani-cli as a fallback

imsamuka:
This PR makes a very simple change: instead of a very sad message box containing `Episode not in library.`, it uses [ani-cli](https://github.com/pystardust/ani-cli/blob/master/ani-cli) as an fallback when no local episode is found. Unfortunately, when the episode is not found by ani-cli, it also returns a very sad `nothing`.

I don't expect it to be merged that much, as it adds a non-python dependency. Presenting and sharing this (working) idea is good enough. 

#600 makes me very excited, ~but i couldn't make it work on my machine yet~ it's working perfectly for now.

When ani-cli fails to find an episode, nothing happens. I couldn't think of a way to get the result back in a good way and alert the user.

Ideally, I think giving the user a list of online sources (builtin or customized, like hooks) would be very cool, but probably would also completely **BLOW UP** trackma project scope.

Also, with a bit of modification, probably is very easy to get fallbacks from `config.json` the same way i made it, if you all don't mind `eval()` shenanigans.

What do you all think about this?

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -945,7 +945,7 @@ class Engine:
 
         if show:
             playing_next = False
-            if not playep:
+            if playep <= 0:
                 playep = show['my_progress'] + 1
                 playing_next = True
 
@@ -954,8 +954,11 @@ class Engine:
 
             self.msg.info(self.name, "Getting %s %s from library..." %
                           (show['title'], playep))
-            filename = self.get_episode_path(show, playep)
-            endep = playep
+
+            try:
+                filename = self.get_episode_path(show, playep)
+            except utils.EngineError:
+                filename = None
 
             if filename:
                 self.msg.info(self.name, 'Found. Starting player...')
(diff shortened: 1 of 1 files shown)


z411:
I'm not sure if we should add such a dependency of a program we're not associated with. Maybe this could be done in a hook.

imsamuka:
I think a hook is a good solution, clean and customizable. An hook like `episode_not_found(engine, show, episode)` can easily reproduce this PR. But I don't know exactly how/where to implement it, since `Engine.play_episode()` actually just return args, instead of playing the episode directly. Is there any good reasons to keep it that way? 

The only drawback are exceptions. Exceptions are treated differently on hooks. Raising an `utils.EngineError` directly on `engine.py` opens an message box (on GTK version), but on hooks, it's just logged. Shouldn't `utils.TrackmaError` and `utils.TrackmaFatal` be treated the same way, when raised on hooks? Is this really intended behavior?

PR #611, opened 2022-05-03
Display icon in taskbar on windows

YUKI2eN3e:


trackma/ui/qt/__init__.py
```diff
@@ -73,6 +73,10 @@ def main(force_qt4=False):
     app = QApplication(sys.argv)
     app.setApplicationName("trackma")
     app.setDesktopFileName("trackma")
+    if os.name == "nt":
+        import ctypes
+        myappid = u'trackma' + utils.VERSION
```
FichteFoll, line 78: Python 2 isn't supported by trackma, so you can drop the `u`.

```suggestion
        myappid = 'trackma' + utils.VERSION
```

trackma/ui/qt/mainwindow.py
```diff
@@ -16,6 +16,7 @@
 
 from trackma import utils
 from trackma import messenger
+import trackma
```
FichteFoll, line 19: This import isn't used.
```suggestion
```

FichteFoll (changes requested):
I cannot test this since I don't use Windows (more than I need to anyway).

PR #618, opened 2022-06-15
[Qt] fix account editing

Birdulon:
When the extra field was added a few years back, it was only added to account creation, not updating. Since tokens expire for some sites, the following error would occur when someone went to update their token after their syncs started getting rejected:
image


FichteFoll (approved):

FichteFoll:
Thanks, I confirmed the fix.

PR #619, opened 2022-06-30
Add more debug logging for episode redirection

FichteFoll:
I added these while debugging why a certain episode wasn't being detected. Turns out that was because my anime relations were outdated (different issue).


FichteFoll:
Merging after a lack of review in order not to have the changes sit idle for too long. I've been using these changes since I made them.

PR #634, opened 2022-08-07
Simplify messenger calls

imsamuka:
I changed the Messenger class to hold a internal class name. Every other class that uses the messenger (`Engine`, `Data`, `Lib` and `TrackerBase`), will instance the messenger class with their `self.name` only once, and every call to this instance, will optionally re-use the given name.

I say optionally, because `debug`, `info`, `warn` and `exception` can still receive a class name, so Hooks continue exactly the same: that's not a breaking change.

I searched and removed `self.name` from every `debug`, `info`, `warn` and `exception` call in the code base. As a bonus, almost every two-line call was inlined.

```
-     self.msg.warn(
-         self.name, "Unknown operation in queue (%s), skipping..." % repr(operation))
+     self.msg.warn("Unknown operation in queue (%s), skipping..." % repr(operation))
```

The code is now simpler, and `self.name` is only used twice in each of those classes.

I revised the changes thrice, but feel free review yourself :smiley:

### DIFF
--- a/trackma/messenger.py
+++ b/trackma/messenger.py
@@ -14,6 +14,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+import traceback
+
 TYPE_DEBUG = 1
 TYPE_INFO = 2
 # TYPE_ERROR = 3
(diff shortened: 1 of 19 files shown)


trackma/messenger.py
```diff
@@ -24,27 +26,40 @@
...
         if self._handler:
-            self._handler(classname, TYPE_INFO, msg)
+            cn, msg = self._parse_msgs(msgs)
+            self._handler(cn, msg_type, msg)
 
-    def warn(self, classname, msg):
-        if self._handler:
-            self._handler(classname, TYPE_WARN, msg)
+    def _parse_msgs(self, msgs):
+        if len(msgs) >= 2:
```
FichteFoll, line 45: Are there any instances left of the log functions being called with an explicit classname? Either way, to simplify the methods here, I'd rather design their signature to represent the 90% use case and add special invocation options for the other, such as a `name` kwarg. Especially with `exception` using a differnt signature already.

FichteFoll (changes requested):
These changes are definitely useful. While I'd prefer to use the stdlib `logging` module in the future at some point, this change certainly doesn't prevent us from going in that direction and in fact it can even serve as the first step.

I just have an issue with the complexity of the call signatures ofthe Message methods now, but after that is handled, you have my approval.

trackma/messenger.py
imsamuka: > Are there any instances left of the log functions being called with an explicit classname?

Yes, the hooks. Including the example in [Hooks Doc](docs/hooks.rst). I thought that making these break would be quite bad, but i guess the crypt function signatures are even worse, i am fixing it.

Since these will be breaking changes, trackma version should bump up, probably?

Also, the `name` kwarg should be mandatory or it can be a positional argument?

```py
def debug(self, msg, name=None): pass # positional or as kword
# or 
def debug(self, msg, *, name=None): pass # only as kword
```

imsamuka:
I thought the same regarding the `logging` module, i feel like we are reinventing the wheel here. But well, its working :+1:.

I changed the function signatures, now it's more obvious. I could make type annotations too, but the parameters explain themselves.
```py
def debug(self, msg, *, name=None):
```
Since it is uncommon, the `name` can only be passed as a kword. It will give us more flexibility changing the methods in the future. If you need to use the same name multiple times, you should use the `with_classname` messenger method before anyway.

FichteFoll (approved):
Thanks, this is much better. I'll leave it to @z411 to decide whether breaking the hook API is acceptable here or not.

z411:
I think breaking the hook API is fine; but I'd prefer to break it only once. If we break it this time, and we later move to the logging module, we'll break it twice. While I appreciate your work, I think if we plan to move to the logging module we should do that now instead.

imsamuka:
I see, so i will revert the last commit. The function signatures will become harder to follow, but the API won't break right now, and we can change to the logging module with more ease.

imsamuka:
Done. The hooks continue the same :+1:

FichteFoll (approved):

z411:
Thank you!

PR #635, opened 2022-08-08
[GTK] fix Gdk ImportError

imsamuka:
This is a very strange bug. When Gdk is loaded before Gtk, the program crashes with:

```
ImportError: Requiring namespace 'Gdk' version '3.0', but '4.0' is already loaded
```

This is easy to fix, but note that this bug was introduced when formatting the code. I did my best to make this as explicit as possible in the code itself, so if this bug is introduced again, it should be more straightforward to fix. I also tested `ypaf`, `black` and `autopep8`, and none of those removed my "fix" yet.

It would be great if someone could reproduce the bug before merging the PR, to be sure this isn't something only I am experiencing.

### DIFF
--- a/trackma/ui/gtk/window.py
+++ b/trackma/ui/gtk/window.py
@@ -19,7 +19,11 @@ import subprocess
 import sys
 import threading
 
-from gi.repository import GLib, Gdk, Gio, Gtk
+from gi.repository import GLib, Gio, Gtk
+
+# When Gdk is loaded before Gtk, the program crashes:
+# "Requiring namespace 'Gdk' version '3.0', but '4.0' is already loaded"
+from gi.repository import Gdk
 
 from trackma import messenger
 from trackma import utils


FichteFoll:
I suppose following [usort's import block concept](https://usort.readthedocs.io/en/stable/guide.html#import-blocks) is a good idea, then. I don't think black does import sorting and I don't know about yapf.

FichteFoll:
I think we should rather do it like this (accoridng to to [the docs](https://python-gtk-3-tutorial.readthedocs.io/en/latest/introduction.html)):

```py
import gi

gi.require_version("Gtk", "4.0")
from gi.repository import Gtk
```

Unless we intend to support gtk4 and 3 with the same code, I guess.

imsamuka:
> I suppose following usort's import block concept is a good idea, then

So should i just place the 2 imports separated by a newline? simple enough

> I think we should rather do it like this (accoridng to to this tutorial):

`gi.require_version` is already being called in [application.py](trackma/ui/gtk/application.py), which is always loaded before `window.py`, i guess. I also made i test here, and even using `gi.require_version` *again* in `window.py`, the ordering still cause a crash

> Unless we intend to support gtk4 and 3 with the same code, I guess.

I also tried changing to `4.0`, and the program crashes, so i guess currently this isn't possible

FichteFoll:
>      I also tried changing to 4.0, and the program crashes, so i guess currently this isn't possible

Yeah, I didn't read the error message properly. that should be `"3.0"`.

Kcchouette:
I've put the following code in [trackma/ui/gtk/\_\_init\_\_.py](https://github.com/z411/trackma/blob/3610b9d3cec6b667432a1d7cbfb4eec2954c3d74/trackma/ui/gtk/__init__.py)

```
import gi
gi.require_version('Gdk', '3.0')
```

FichteFoll:
I tried around a few things and came to the conclusion that the error occurs if you import either of Gdk or Gtk before setting their respective version and then import the other. The solution appears to be to just require them both, which is what I did in 16c1551916ae5567cb2635cb0d1bfb5de9373957.

Issue #628, opened 2022-07-12
Drop Qt4 and support Qt6


txtsd:
Qt4 has been EOL since 2015.

It's time to drop support in the codebase, turn Qt5 into legacy, and focus on supporting Qt6 instead.

alsoGAMER:
i was working on this on my fork [here](https://github.com/alsoGAMER/trackma/tree/qt6-rework), if anyone could contribute that would help speeding up the process

txtsd:
Oh good. I'll test it later today or tomorrow and see how I can help.

alsoGAMER:
> Oh good. I'll test it later today or tomorrow and see how I can help.

thanks!
the state of that fork is mostly working; it starts, renders the lists (though without the colored bars) and images, also updating the lists locally and remotely works fine).

there are still some issues with writing to files (especially the config one), but that is something I'm unable to work out how to fix.

txtsd:
@alsoGAMER How are you launching it? `trackma-qt -6` does not do anything different, and it seems to be using qt5 like normal.

alsoGAMER:
@txtsd make sure your local clone isn't broken, i made some big changes to the repo some days ago.

also I'm using an archlinux [distrobox](https://github.com/89luca89/distrobox) to run it with qt6 and it works flawlessly, so you can try that.

alsoGAMER:
@txtsd did you get it working?

txtsd:
Not yet. I've been battling respiratory illnesses :(

alsoGAMER:
> Not yet. I've been battling respiratory illnesses :(

Oh i see, hope you'll recover soon.

alsoGAMER:
I don't know why is behaving that way, i just retried the setup from scratch and worked perfectly fine.

![image](https://user-images.githubusercontent.com/35269770/208128190-16ebcc61-c9ae-45bf-bf93-2ff4ba2f61fa.png)
![image](https://user-images.githubusercontent.com/35269770/208128464-3b812189-10f4-4e24-be22-6f7125dfd871.png)

v-fox:
Rebased @alsoGAMER's patch to get rid of PyQt5 from the system: [trackma_qt6-rework.zip](https://github.com/z411/trackma/files/14830108/trackma_qt6-rework.zip)
For some reason github craps itself on unzipped patch-file.

PR #653, opened 2022-11-01
feat: Add Poetry

### LINKED ISSUE #609: Specifying project dependencies
Currently, there is a list of general dependencies in the readme for various aspects of trackma, but those don't map cleanly to Python package names on pypi. It would seem rather useful to me to have these in some form, either in `setup.py` as extras or the readme.

At the same time, what do you know of `pyproject.toml` and using e.g. poetry for managing development environments instead of `setup.py`? I've had a rather good experience with those and they definitely appear to be the way to go forward with, ~~although poetry's dependency groups feature with its recent additions is currently not released in any version~~.

txtsd:
This adds dependencies for a virtual environment. 
Removes `pyqt4`.
Updates minimum python version.
Removes PyQt4 since Qt4 has been EOL for several years.

~~Python is currently pinned to 3.7.14 because `pycairo` is not compatible with python 3.6.
Let me know if the versions should be `>=` instead of `==`.~~

Should close #609 when done

EDIT: ~~Hold on I might be missing gtk deps.~~ Nevermind.
EDIT: Additionally, drops `setup.py` in favor of `poetry`, and updates the `README` with relevant info.
EDIT: Python is now `3.7+`


pyproject.toml
```diff
@@ -0,0 +1,28 @@
+[tool.poetry]
+name = "trackma"
+version = "2022.10"
+description = "Open multi-site list manager"
+authors = ["z411 <z411@krutt.org>"]
+license = "GPL-3.0-or-later"
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "3.7.14"
```
FichteFoll, line 10: This should definitely be `^3.7`. pycairo should be an optional dependency, but [3.6 has reached EOL by almost a year now](https://peps.python.org/pep-0494/) so requiring 3.7 should be fine. That also allows us to use `from __future__ import annotations` and makes way for a lot of useful annotation features that have been added in later versions.

pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+pygobject = "3.42.1"
+pyinotify = "0.9.6"
+pypresence = "4.2.1"
+pyqt5 = "5.15.7"
+requests = "2.28.1"
+requests-oauthlib = "1.3.1"
+twitter = "1.19.3"
+urwid = "2.1.2"
+setuptools = "63.1.0"
+sip = "6.6.2"
```
FichteFoll, line 24: The UI-specific dependencies should be optional and specified with extras.

See https://python-poetry.org/docs/pyproject/#extras

setup.py
```diff
@@ -42,16 +42,18 @@
         'gui_scripts': [
             'trackma-gtk = trackma.ui.gtk:main [GTK]',
             'trackma-qt = trackma.ui.qt:main [Qt]',
-            'trackma-qt4 = trackma.ui.qt.qt4:main [Qt4]',
         ]
     },
```
FichteFoll, line 46: Please add these scripts to the poetry file as well.

https://python-poetry.org/docs/pyproject/#scripts

README.md
```diff
@@ -49,7 +49,7 @@ Dependencies
 
 The only required dependency to run Trackma is:
 
-- Python 3.4/3.5
+- Python 3.7.14
```
FichteFoll, line 52:
```suggestion
- Python 3.7+
```

pyproject.toml
```diff
@@ -0,0 +1,28 @@
+[tool.poetry]
+name = "trackma"
+version = "2022.10"
```
FichteFoll, line 3: trackma uses a different versioning scheme.

FichteFoll (changes requested):

pyproject.toml
txtsd: Done

setup.py
txtsd: Done

README.md
txtsd: Done

FichteFoll (approved):
I haven't tested thislocally but the changes look good to me.

txtsd:
Conflicts resolved.

HS-157 (approved):
Hi, I am not a expert with Poetry but I tested locally following the instruction from the Readme.
And it's run with differently frontend on Archlinux !
```
❯ python --version
Python 3.10.8
❯ poetry --version
Poetry (version 1.3.1)
```
For me, all it's good, thank you.

```
❯ poetry run trackma-gtk
Trackma GTK v0.8.5
Engine: Trackma v0.8.5 - using account HS157(anilist).
Data: Initializing...
libanilist: Initializing...
Data: Using libanilist (anime)
libanilist: Downloading list...
Engine: Parsing redirection file...
Engine: Scanning local library...
Tracker (pyinotify): Initializing...
Tracker (pyinotify): Using pyinotify.
Ready.
Engine: Unloading...
Tracker (pyinotify): Unloading...
Engine: Unloading user hooks...
```
2022-12-16-012502_grim

txtsd:
Wow PEP 582 is YUM!

z411:
Does it support to load the program version from `utils.VERSION`? I like to keep the version in one place as to avoid inconsistencies.

txtsd:
> Does it support to load the program version from `utils.VERSION`? I like to keep the version in one place as to avoid inconsistencies.

I don't understand. Can you elaborate on what you mean?

txtsd:
@z411 There is [poetry-dynamic-versioning](https://pypi.org/project/poetry-dynamic-versioning/) which can set the version, but it only grabs the version from tags. So unless you want to make tags the true source of the version.

If we go that way, it will also require running `poetry dynamic-versioning` (note the space) to set the value in `pyproject.toml` and `utils.py` so they can then be committed. This is because by default, `poetry-dynamic-versioning` sets the version while it's working/building (ideally to a placeholder (`version = "0.0.0"`), then reverts it when it's done.

Another solution is to use a pre-commit hook to block when the version values in `pyproject.toml` and `utils.py` are different.

FichteFoll:
I thought the dynamic versioning plugin and the pre-commit hook were two different approaches to this problem? Imo this can also be done at a later point. I'd merge the PR as soon as I find some time in my day to test the various poetry commands on it locally, but it does look good already.

txtsd:
> I thought the dynamic versioning plugin and the pre-commit hook were two different approaches to this problem? Imo this can also be done at a later point.

Yes, just a little holiday haze 😅

> I'll merge the PR as soon as I find some time in my day to test the various poetry commands on it locally, but it does look good already.

Okay good! I'll look into getting Qt6 support for #628 after that.

FichteFoll:
Finally managed to test this properly (after I fixed the problems I had with poetry&pip always trying to use some keyring that I never configured, https://github.com/pypa/pip/issues/6773, https://github.com/jaraco/keyring#disabling-keyring).

Also ran into an issue with Arch packaging regarding poetry and had to manually downgrade a package there (https://bugs.archlinux.org/task/77375).

I have made some slight adjustments to the readme and merged it into `master`. Note that the readme now does not correctly reflect the extras included on the latest version on pypi until the next version update.

txtsd:
Awesome!

PR #654, opened 2022-11-02
Set desktop file name for Wayland icon

KodyVB:
I found this solution to Wayland not showing the icon for PyQt applications thanks to the KDE guidelines [here](https://community.kde.org/Guidelines_and_HOWTOs/Wayland_Porting_Notes#Application_Icon). 

I was having the same issue of the Wayland yellow and white icon showing in the application's title bar and in the task manager on the panel with some PyQt6 applications I made. When I added this line, it fixed the icon issue for both locations with my applications, but for some reason it's not fixing it for Trackma's title bar icon. Still, I figured this was better than nothing, as it fixes it for the task manager on the panel.

Edit:  I hadn't realized setDesktopFileName was already called in \_\_init\_\_.py. The second commit uses that original call instead of calling the function again. Still having the same problem of it not fixing the title bar in KDE Wayland, though.

### DIFF
--- a/trackma/ui/qt/__init__.py
+++ b/trackma/ui/qt/__init__.py
@@ -73,7 +73,7 @@ def main(force_qt4=False):
 
     app = QApplication(sys.argv)
     app.setApplicationName("trackma")
-    app.setDesktopFileName("Trackma-qt")
+    app.setDesktopFileName("trackma-qt")
     if os.name == "nt":
         import ctypes
         myappid = 'trackma' + utils.VERSION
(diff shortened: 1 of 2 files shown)


KodyVB:
Okay, making the "T" in trackma-qt lowercase made it work in both locations. I don't know why I didn't think of trying this earlier. I honestly feel a bit silly about it and wish I would have thought about it before that first or second commit, lol. This seems to be working just fine for me on KDE's Wayland.

alsoGAMER:
Can confirm this fixes the issue

FichteFoll (approved):

Issue #660, opened 2022-12-09
Add the option to "list" in a different format


RaitaroH:
Hi,

Thank you for the awesome program. I am heavily using it with my adl script.

I think it would be very helpful, both to me for my adl script, but also others like [justchokingaround](https://github.com/justchokingaround)/[trackma-wrapper](https://github.com/justchokingaround/trackma-wrapper) to allow `trackma -a 1 list` to output in a different format upon request. What I mean is, by default it shows a table, but for scripting purposes having it in a character delimited format, or even json would be much better.

For example, right now the output looks like this if I copy paste it:
```
| Index  Title                                           Progress  Score  |
|   1    [0;34mBlue Lock[0m......................................   9 / 24   5.5   |
|   2    [0;34mBocchi the Rock![0m...............................   9 / 12   6.5   |
|   3    [0;34mChainsaw Man[0m...................................   9 / 12   5.5   |
|   4    [0;34mDo It Yourself!![0m...............................  10 / 12    6    |
|   5    [0;31mIsekai Oji-san[0m.................................   9 / 13   6.5   |
|   6    Kyokou Suiri...................................   4 / 12    0    |
|   7    [0;34mMairimashita! Iruma-kun 3[0m......................   9 / 21    5    |
|   8    [0;34mMeitantei Conan[0m................................ 1066 / ?   7.5   |
|   9    [0;34mSPY×FAMILY Part 2[0m..............................  10 / 13    7    |
|   10   [0;34mTensei Shitara Ken Deshita[0m.....................  11 / 12    5    |
|   11   [0;31mUchi no Shishou wa Shippo ga Nai[0m...............  10 / 13   6.5   |
11 results
````

This requires a bit of scripting to clean up and to get anime titles, or to use `trackma info index` which is slow.

As such I believe that following would be much nicer: `trackma -a 1 list raw`
```
1|"Blue Lock"|9|24|5.5
2|"Bocchi the Rock!"|9|12|6.5
```
This allows for very simple `awk -F'|' '{print $n}'` type commands.

FichteFoll:
Separating the fields with a tab (or using JSON) would be a better approach than a `|`.

RaitaroH:
@FichteFoll I made a PR for this. Hopefully it will get accepted.

Issue #659, opened 2022-12-09
Indicate when you are behind watching an ongoing anime [feature request]


RaitaroH:
At the moment anilist has a calendar view where I can see what shows I am behind on:
![image](https://user-images.githubusercontent.com/19552720/206740291-de265afb-7daa-402b-9ecf-c6f6d049e720.png)

It would be really nice to have some indicators of this information in trackma.
I suppose an easy way of doing it is just to color the titles diffrently like so: 
![image](https://user-images.githubusercontent.com/19552720/206740930-57628fd6-9ec4-4b70-97e1-3c98617982b5.png)

A more complicated calendar view can be found in justfoolingaround/animdl, if this is useful.

I am particularly interested in this for the cli variant of trackma.

Note: to produce the img, I have sent the `trackma -a 1 list` output to a file, changed `0;34m` to `0;31m` to be red and took a screenshot.

z411:
I also like this feature of Anilist. It'd be worth implementing.

z411:
Implemented it in the latest commit, see if it works.

![image](https://user-images.githubusercontent.com/1988512/206878907-86a9748c-7394-4f18-a669-3fdfc8f8800d.png)

It needs to be done for the rest of the CLIs, but the logic is very easy as the function to get the number of aired episodes is already there.

RaitaroH:
@z411 I can confirm the latest commit works (pip).
Thank you for the quick implementation.

PR #662, opened 2023-01-04
list shows in json format

RaitaroH:
For issue https://github.com/z411/trackma/issues/660
If you merge this PR, please squash and merge.

```
$ trackma help
...
     list|ls <>            Lists all shows available in the local list.
   list_json <>            Print list of shows in json format.
...
```
Example of `trackma list_json` (it defaults to watching list)
```json
{"title": "Ars no Kyojuu", "current_episode": "2", "total_episodes": "12", "score": "0", "status": "Ongoing", "estimated_aired_episode": "2", "image": "https://s4.anilist.co/file/anilistcdn/media/anime/cover/medium/bx155089-I5zYFNvWua57.png"}
{"title": "Blue Lock", "current_episode": "14", "total_episodes": "24", "score": "5.5", "status": "Ongoing", "estimated_aired_episode": "14", "image": "https://s4.anilist.co/file/anilistcdn/media/anime/cover/medium/bx137822-4dVWMSHLpGf8.png"}
{"title": "Buddy Daddies", "current_episode": "2", "total_episodes": "13", "score": "0", "status": "Ongoing", "estimated_aired_episode": "2", "image": "https://s4.anilist.co/file/anilistcdn/media/anime/cover/medium/bx155907-gR7aRwVHwrjc.jpg"}
{"title": "Bungou Stray Dogs 4th Season", "current_episode": "2", "total_episodes": "13", "score": "0", "status": "Ongoing", "estimated_aired_episode": "2", "image": "https://s4.anilist.co/file/anilistcdn/media/anime/cover/medium/bx141249-8tjavEDHmLoT.jpg"}
{"title": "Flying Witch", "current_episode": "12", "total_episodes": "12", "score": "0", "status": "Finished", "image": "https://s4.anilist.co/file/anilistcdn/media/anime/cover/medium/bx21284-vQcCLIWt1o5O.png"}
```
Then it also can accept list status, `trackma list_json paused`:
```json
{"title": "Casshern Sins", "current_episode": "15", "total_episodes": "24", "score": "6", "status": "Finished", "image": "https://s4.anilist.co/file/anilistcdn/media/anime/cover/medium/bx4981-V6MbMiJUqdvP.jpg"}
{"title": "Chi's Sweet Home: Atarashii Ouchi", "current_episode": "63", "total_episodes": "104", "score": "0", "status": "Finished", "image": "https://s4.anilist.co/file/anilistcdn/media/anime/cover/medium/bx6024-eYiQMRfPfWiF.jpg"}
{"title": "Joshiraku", "current_episode": "1", "total_episodes": "13", "score": "0", "status": "Finished", "image": "https://s4.anilist.co/file/anilistcdn/media/anime/cover/medium/bx12679-0zlvLP7KyxjH.png"}
{"title": "Karakai Jouzu no Takagi-san 2", "current_episode": "7", "total_episodes": "12", "score": "5", "status": "Finished", "image": "https://s4.anilist.co/file/anilistcdn/media/anime/cover/medium/bx107068-KJq0eFP0GTjL.jpg"}
```
I believe `trackma -a` should work. I tested with `trackma -a 2 list_json` and it asks me to select account first, before listing anything.

In bash scripts, `jq` works as expected. Here is a more complicated command:
```bash
trackma list_json | jq -jr '.title, "|", .current_episode, "|", .total_episodes, "|", .score, "|", .status, "|",.estimated_aired_episode, "\n"'
```
```
Ars no Kyojuu|2|12|0|Ongoing|2
Blue Lock|13|24|5.5|Ongoing|14
Buddy Daddies|2|13|0|Ongoing|2
Bungou Stray Dogs 4th Season|2|13|0|Ongoing|2
Flying Witch|9|12|0|Finished|null
```

### DIFF
--- a/trackma/ui/cli.py
+++ b/trackma/ui/cli.py
@@ -991,7 +991,7 @@ class Trackma_cmd(command.Cmd):
             j = {
                 "title": title_str,
                 "current_episode": episode_str_current,
-                "final_episode": episode_str_last,
+                "total_episodes": episode_str_last,
                 "score": score
             }
 
--- a/trackma/ui/cli.py
+++ b/trackma/ui/cli.py
@@ -991,7 +991,7 @@ class Trackma_cmd(command.Cmd):
             # Check if show is airing and get estimated aired episode
             if show['status'] == utils.Status.AIRING:
                 estimate = utils.estimate_aired_episodes(show)
-                j["estimated_aired_episode"] = estimate
+                j["estimated_aired_episode"] = "{0}".format(estimate)
 
             print(json.dumps(j))
 
--- a/trackma/ui/cli.py
+++ b/trackma/ui/cli.py
@@ -75,7 +75,6 @@ class Trackma_cmd(command.Cmd):
         'sort':         1,
         'mediatype':    (0, 1),
         'info':         1,
-        'name':         1,
         'search':       1,
         'list_json':    (0, 1),
         'add':          1,
@@ -382,21 +381,6 @@ class Trackma_cmd(command.Cmd):
                 return
 
         self._make_list_json(self.sortedlist)
-
-    def do_name(self, args):
-        """
-        Print title name for given index.
-
-        :param show Show index.
-        :usage name <index>
-        """
-        try:
-            show = self._get_show(args[0])
-        except utils.TrackmaError as e:
-            self.display_error(e)
-            return
-
-        print(show['title'])
 
     def do_info(self, args):
         """
--- a/trackma/ui/cli.py
+++ b/trackma/ui/cli.py
@@ -22,6 +22,7 @@ import shlex
 import subprocess
 import sys
 import textwrap
+import json
 from operator import itemgetter  # Used for sorting list
 
 from trackma import messenger
@@ -76,6 +77,7 @@ class Trackma_cmd(command.Cmd):
         'info':         1,
         'name':         1,
         'search':       1,
+        'list_json':    (0, 1),
         'add':          1,
         'del':          1,
         'delete':       1,
@@ -363,12 +365,23 @@ class Trackma_cmd(command.Cmd):
         # Show the list in memory
         self._make_list(self.sortedlist)
 
-    def do_list_raw(self, args):
+    def do_list_json(self, args):
         """
-        Similar to list except remove formatting.
-        : name list_raw
+        Print list of shows in json format.
+        Use the command `filter` without arguments to see the available statuses.
+
+        :name list_json
+        :usage list_json <status>
         """
-        self._make_list_raw(self.sortedlist)
+        if args:
+            try:
+                self.filter_num = self._guess_status(args[0].lower())
+                self._load_list()
+            except KeyError:
+                print("Invalid filter.")
+                return
+
+        self._make_list_json(self.sortedlist)
 
     def do_name(self, args):
         """
@@ -379,7 +392,6 @@ class Trackma_cmd(command.Cmd):
         """
         try:
             show = self._get_show(args[0])
-            details = self.engine.get_show_details(show)
         except utils.TrackmaError as e:
             self.display_error(e)
             return
@@ -969,26 +981,48 @@ class Trackma_cmd(command.Cmd):
         print('%d results' % len(showlist))
         print()
 
-    def _make_list_raw(self, showlist):
+    def _make_list_json(self, showlist):
         """
-        Helper function for printing a non-formatted show list
+        helper function for printing a json formatted show list
         """
         altnames = self.engine.altnames()
 
-        # List shows
+        episode_str_current = ""
+        episode_str_last = ""
+        # list shows
         for index, show in showlist:
             if self.engine.mediainfo['has_progress']:
-                episodes_str = "{0}\t{1}".format(
-                    show['my_progress'], show['total'] or '?')
-            else:
-                episodes_str = "-"
+                episode_str_current = "{0}".format(show['my_progress'] or '0')
+                episode_str_last = "{0}".format(show['total'] or '?')
 
-            # Get title (and alt. title) and if need be, truncate it
+            # get title (and alt. title) and if need be, truncate it
             title_str = show['title']
             if altnames.get(show['id']):
                 title_str += " [{}]".format(altnames.get(show['id']))
 
-            print(index,title_str, episodes_str,show['my_score'],sep='\t')
+            # ensure score is string; anilist can have string scores such as stars or smiles
+            score = "{0}".format(show['my_score'])
+
+            # json dictionary
+            j = {
+                "title": title_str,
+                "current_episode": episode_str_current,
+                "final_episode": episode_str_last,
+                "score": score
+            }
+
+            # Color title according to status
+            if show['status'] == utils.Status.AIRING:
+                estimate = utils.estimate_aired_episodes(show)
+                if estimate and show['my_progress'] < estimate:
+                    # User is behind the (estimated) aired episode
+                    j["color"] = _COLOR_BEHIND
+                else:
+                    j["color"] = _COLOR_AIRING
+            else:
+                j["color"] = _COLOR_RESET
+
+            print(json.dumps(j))
 
 class Trackma_accounts(AccountManager):
     def _get_id(self, index):
@@ -1027,12 +1061,12 @@ class Trackma_accounts(AccountManager):
                     password = getpass.getpass('Enter password (no echo): ')
                 elif selected_api[2] in [utils.Login.OAUTH, utils.Login.OAUTH_PKCE]:
                     username = input('Enter account name: ')
-                    
+
                     auth_url = selected_api[3]
                     if selected_api[2] == utils.Login.OAUTH_PKCE:
                         extra['code_verifier'] = utils.oauth_generate_pkce()
                         auth_url = auth_url % extra['code_verifier']
-                    
+
                     print('OAuth Authentication')
                     print('--------------------')
                     print('This website requires OAuth authentication.')
--- a/trackma/ui/cli.py
+++ b/trackma/ui/cli.py
@@ -362,6 +362,13 @@ class Trackma_cmd(command.Cmd):
         # Show the list in memory
         self._make_list(self.sortedlist)
 
+    def do_list_raw(self, args):
+        """
+        Similar to list except remove formatting.
+        : name list_raw
+        """
+        self._make_list_raw(self.sortedlist)
+
     def do_info(self, args):
         """
         Gets detailed information about a local show.
@@ -945,6 +952,26 @@ class Trackma_cmd(command.Cmd):
         print('%d results' % len(showlist))
         print()
 
+    def _make_list_raw(self, showlist):
+        """
+        Helper function for printing a non-formatted show list
+        """
+        altnames = self.engine.altnames()
+
+        # List shows
+        for index, show in showlist:
+            if self.engine.mediainfo['has_progress']:
+                episodes_str = "{0}\t{1}".format(
+                    show['my_progress'], show['total'] or '?')
+            else:
+                episodes_str = "-"
+
+            # Get title (and alt. title) and if need be, truncate it
+            title_str = show['title']
+            if altnames.get(show['id']):
+                title_str += " [{}]".format(altnames.get(show['id']))
+
+            print(index,title_str, episodes_str,show['my_score'],sep='\t')
 
 class Trackma_accounts(AccountManager):
     def _get_id(self, index):
--- a/trackma/ui/cli.py
+++ b/trackma/ui/cli.py
@@ -979,10 +979,8 @@ class Trackma_cmd(command.Cmd):
                 episode_str_current = "{0}".format(show['my_progress'] or '0')
                 episode_str_last = "{0}".format(show['total'] or '?')
 
-            # get title (and alt. title) and if need be, truncate it
+            # get title
             title_str = show['title']
-            if altnames.get(show['id']):
-                title_str += " [{}]".format(altnames.get(show['id']))
 
             # ensure score is string; anilist can have string scores such as stars or smiles
             score = "{0}".format(show['my_score'])
--- a/trackma/ui/cli.py
+++ b/trackma/ui/cli.py
@@ -969,41 +969,29 @@ class Trackma_cmd(command.Cmd):
         """
         helper function for printing a json formatted show list
         """
-        altnames = self.engine.altnames()
 
-        episode_str_current = ""
-        episode_str_last = ""
         # list shows
         for index, show in showlist:
             if self.engine.mediainfo['has_progress']:
                 episode_str_current = "{0}".format(show['my_progress'] or '0')
                 episode_str_last = "{0}".format(show['total'] or '?')
 
-            # get title
-            title_str = show['title']
-
             # ensure score is string; anilist can have string scores such as stars or smiles
             score = "{0}".format(show['my_score'])
 
             # json dictionary
             j = {
-                "title": title_str,
+                "title": show['title'],
                 "current_episode": episode_str_current,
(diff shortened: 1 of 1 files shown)


RaitaroH:
Apologies if this isn't quite correct / good. I don't know python that much.

RaitaroH:
[print name of show only](https://github.com/z411/trackma/pull/662/commits/0b1137e3c9d63df663e07a9ea65f855cfc853cec) commit as `trackma info index` can be slow (at least on my end). This new function is much faster. Would be nice to have; [this](https://github.com/justchokingaround/trackma-wrapper) uses info for example.

Please ignore this commit if you wish. At the very least please consider merging the `list_raw` commit.

trackma/ui/cli.py
```diff
@@ -362,6 +363,29 @@ def do_list(self, args):
...
+    def do_name(self, args):
+        """
+        Print title name for given index.
+
+        :param show Show index.
+        :usage name <index>
+        """
+        try:
+            show = self._get_show(args[0])
+            details = self.engine.get_show_details(show)
```
FichteFoll, line 382: This variable is unused.

trackma/ui/cli.py
```diff
@@ -362,6 +363,29 @@ def do_list(self, args):
         # Show the list in memory
         self._make_list(self.sortedlist)
 
+    def do_list_raw(self, args):
+        """
+        Similar to list except remove formatting.
+        : name list_raw
```
FichteFoll, line 369: This line seems incomplete?

trackma/ui/cli.py
```diff
@@ -945,6 +969,26 @@ def _make_list(self, showlist):
...
+                    show['my_progress'], show['total'] or '?')
+            else:
+                episodes_str = "-"
+
+            # Get title (and alt. title) and if need be, truncate it
+            title_str = show['title']
+            if altnames.get(show['id']):
+                title_str += " [{}]".format(altnames.get(show['id']))
+
+            print(index,title_str, episodes_str,show['my_score'],sep='\t')
```
FichteFoll, line 991: Is the `index` actually something that tools need if they also have the name? I was generally under the impression that the index is primarily useful as a shortcut in interactive usage, but that's for humans and not for tools as they can just copy the proper anime name around. Besides, the index can change easily by any entry being added or removed from the list, making them very volatile and not safe to use for tooling. The unique `name` attribute of an entry should be superior in every way, aside from changing the language preference (english vs romaji for example), but the index would be even less safe then since the entire list would be have been reordered.

Similarly, what is the `name` command supposed to be used for?

And finally, I worry about the extensibility of this command as it currently uses a fixed list of columns, but what if we every wanted to add more to it or remove a column? I mean, even this implementation already has a variable number of columns depending on `has_progress`. 

The solutions I see are

1. Add a header line that names the respective columns
2. Just output JSON instead.

The decision for whether a tab-delimited table or JSON would be easier to use heavily depends on the use case & tooling, but JSON in general is much more flexible for tool-to-tool communication with rich data and frequently used by other tools.

trackma/ui/cli.py
RaitaroH: What about
```python
    def do_list_raw(self, args):
        """
        Print list of shows in json format.

        : name list_raw
        """
        self._make_list_raw(self.sortedlist)
```

trackma/ui/cli.py
RaitaroH: Well, `index` is indeed not needed.
Consider the following json approach:
```python
    def _make_list_raw(self, showlist):
        """
        helper function for printing a json formatted show list
        """
        altnames = self.engine.altnames()

        episode_str_current = ""
        episode_str_last = ""
        # list shows
        for index, show in showlist:
            if self.engine.mediainfo['has_progress']:
                episode_str_current = "{0}".format(show['my_progress'] or '0')
                episode_str_last = "{0}".format(show['total'] or '?')

            # get title (and alt. title) and if need be, truncate it
            title_str = show['title']
            if altnames.get(show['id']):
                title_str += " [{}]".format(altnames.get(show['id']))

            # ensure score is string; anilist can have string scores such as stars or smiles
            score = "{0}".format(show['my_score'])

            # print(index,title_str,episodes_str,show['my_score'],sep='\t')
            j = {
                "title": title_str,
                "current_episode": episode_str_current,
                "final_episode": episode_str_last,
                "score": score
            }
            print(json.dumps(j))

```
which prints:
```
{"title": "Ars no Kyojuu", "current_episode": "1", "final_episode": "12", "score": "0"}
{"title": "Blue Lock", "current_episode": "13", "final_episode": "24", "score": "5.5"}

```
As current_episode is number and final_episode could be string (?), I format both.
Score can also potentially be stars or smiles on anilist, so I am not sure how to deal with that there.

`trackma list_raw | jq -r '.title'` works as expected on my end. Of course, score, final_episode and current_episode also work.

---

As mentioned before `trackma info` can be a bit slow as it has to print more info. `name` would just print the name for a given index. But, you are right the index can change, so is better to ignore it. As I said above feel free to ignore the `name`.

If this json approach is suitable, I can make another PR or change this one.

RaitaroH: After making some changes to my personal adl script, to use trackma list_json, I also propose the following change to my above code, in order to also print the color code.
```python
            # json dictionary
            j = {
                "title": title_str,
                "current_episode": episode_str_current,
                "final_episode": episode_str_last,
                "score": score
            }

            # Color title according to status
            if show['status'] == utils.Status.AIRING:
                estimate = utils.estimate_aired_episodes(show)
                if estimate and show['my_progress'] < estimate:
                    # User is behind the (estimated) aired episode
                    j["color"] = _COLOR_BEHIND
                else:
                    j["color"] = _COLOR_AIRING
            else:
                j["color"] = _COLOR_RESET

            print(json.dumps(j))
```
This prints:
```
{"title": "Ars no Kyojuu", "current_episode": "1", "final_episode": "12", "score": "0", "color": "\u001b[0;34m"}
{"title": "Blue Lock", "current_episode": "13", "final_episode": "24", "score": "5.5", "color": "\u001b[0;34m"}
```
The else with color_reset could be left out, but perhaps this might cause issues in other people's code.
`trackma list_json | jq -r '.color // "color value"'` could be used in bash.

FichteFoll: ~~`I suggest using an enum with string values istead of ANSI escape codes for "airing/behind" status. The color representation of a status is an implementation detail.~~

Actually, I'd put "airing" in a field on its own and the information used for the "behind" field, the estimated latest episode, also in its own field because then tooling can decide for itself how it wants to treat entries like that.

Going forward, we might also be interested in an option that selects the fields to be printed, but that can come in later as it would be backwards-compatible.

trackma/ui/cli.py
```diff
@@ -74,7 +75,9 @@ class Trackma_cmd(command.Cmd):
         'sort':         1,
         'mediatype':    (0, 1),
         'info':         1,
+        'name':         1,
```
FichteFoll, line 78: Please just remove this command if we don't see a use case for it.

trackma/ui/cli.py
```diff
@@ -362,6 +365,39 @@ def do_list(self, args):
         # Show the list in memory
         self._make_list(self.sortedlist)
 
+    def do_list_json(self, args):
+        """
+        Print list of shows in json format.
```
FichteFoll, line 370: Should mention that each entry is printed as a JSON line on its own (instead of an array). At least right now. 

If we were to consider things like the scoring format, it may be useful to also report that in a meta object on its own, in which case the question follows whether the entries should instead be wrapped in an array to build an object as follows:

```json
{ "account": "string", "api": "string", "has_progress": false, "scoring_format": "enum", "entries": [{…}] }
```

I don't now how the scoring format options in anilist are implemented currently in various parts of the code or whether it makes more sense to instead specify the `max_score` to be of either 3, 5, or 10 (and use the latter for 10, 20, and 100-point schemas).

And that it is intended to be used by third-party tools.

trackma/ui/cli.py
```diff
@@ -945,6 +981,48 @@ def _make_list(self, showlist):
...
+        # list shows
+        for index, show in showlist:
+            if self.engine.mediainfo['has_progress']:
+                episode_str_current = "{0}".format(show['my_progress'] or '0')
+                episode_str_last = "{0}".format(show['total'] or '?')
+
+            # get title (and alt. title) and if need be, truncate it
+            title_str = show['title']
+            if altnames.get(show['id']):
+                title_str += " [{}]".format(altnames.get(show['id']))
```
FichteFoll, line 1001: Better use an `altname` field than modify the `title`.

trackma/ui/cli.py
```diff
@@ -945,6 +981,48 @@ def _make_list(self, showlist):
...
+
+            # ensure score is string; anilist can have string scores such as stars or smiles
+            score = "{0}".format(show['my_score'])
+
+            # json dictionary
+            j = {
+                "title": title_str,
+                "current_episode": episode_str_current,
+                "final_episode": episode_str_last,
```
FichteFoll, line 1010:
```suggestion
                "total_episodes": episode_str_last,
```

FichteFoll (changes requested):

trackma/ui/cli.py
RaitaroH: For my particular usage, each entry being on its own line is better. In case another shell program wants to use this, I think it might be better to leave it like this.

The scoring format is something I was simply wondering about, but I think is a good idea to investigate.

trackma/ui/cli.py
RaitaroH: I kept the code from _make_list function.
I think just having: `title_str = show['title']` should be enough; otherwise I am not sure of a better implementation using altnames.

FichteFoll: If you don't think the altname is useful (for you), you can just skip it). Either way, you should not alter the `title` here.

RaitaroH: Resolved. Simply `title_str = show['title']` and pass that to json.

trackma/ui/cli.py
RaitaroH: Good idea. I made changes for this.

```python
            # json dictionary
            j = {
                "title": title_str,
                "current_episode": episode_str_current,
                "total_episodes": episode_str_last,
                "score": score,
                "airing_status": False,
                "behind_status": False
            }

            # Check if show is airing and if user is behind
            if show['status'] == utils.Status.AIRING:
                estimate = utils.estimate_aired_episodes(show)
                if estimate and show['my_progress'] < estimate:
                    # User is behind the (estimated) aired episode
                    j["airing_status"] = True;
                    j["behind_status"] = True;
                else:
                    j["airing_status"] = True;

            print(json.dumps(j))
```
I have tested on my end and prints as expected.
```json
{"title": "Saikyou Onmyouji no Isekai Tenseiki", "current_episode": "1", "total_episodes": "?", "score": "0", "airing_status": true, "behind_status": false}
{"title": "Scrapped Princess", "current_episode": "2", "total_episodes": "24", "score": "0", "airing_status": false, "behind_status": false}
{"title": "Sugar Apple Fairy Tale", "current_episode": "1", "total_episodes": "12", "score": "0", "airing_status": true, "behind_status": true}
{"title": "TRIGUN STAMPEDE", "current_episode": "1", "total_episodes": "?", "score": "0", "airing_status": true, "behind_status": false}
```

trackma/ui/cli.py
```diff
@@ -990,19 +990,20 @@ def _make_list_json(self, showlist):
...
             }
 
-            # Color title according to status
+            # Check if show is airing and if user is behind
             if show['status'] == utils.Status.AIRING:
                 estimate = utils.estimate_aired_episodes(show)
                 if estimate and show['my_progress'] < estimate:
                     # User is behind the (estimated) aired episode
-                    j["color"] = _COLOR_BEHIND
+                    j["airing_status"] = True;
```
FichteFoll, line 1003: Actually, why not just dump `show['status']` here?

trackma/ui/cli.py
FichteFoll: :+1: Let's close this in favor of more focues discussions.

trackma/ui/cli.py
```diff
@@ -990,19 +990,20 @@ def _make_list_json(self, showlist):
...
+                    j["behind_status"] = True;
```
FichteFoll, line 1004: Still not too fond of this boolean (and the `status` in the field name). why not report the estimated last aired episode as-is?

trackma/ui/cli.py
RaitaroH: Well, if I do that I get:
`TypeError: Object of type Status is not JSON serializable`
I also tried to simply have a variable named status and no if else statement. Same issue.
But if I simply print status it just says "Ongoing".

Perhaps I just didn't use it right?

trackma/ui/cli.py
RaitaroH: Hmm, this was as per 9cfc7d069dfe1e8daa8f703cd636c9c4dcb084ed as requested by myself in #659.
You are suggesting to print the last aired episode, and let scripts calculate if the user is caught up or behind?
Wouldn't it be confusing when "current_episode" exists?

trackma/ui/cli.py
FichteFoll: Yeah, you'll need to use `show['status'].value`.

trackma/ui/cli.py
FichteFoll: > Hmm, this was as per [9cfc7d0](https://github.com/z411/trackma/commit/9cfc7d069dfe1e8daa8f703cd636c9c4dcb084ed) as requested by myself in #659.

Not sure how this is relevant since that code is specifically about the CLI representation in the table for the user, not for tools.

> You are suggesting to print the last aired episode, and let scripts calculate if the user is caught up or behind?

If that is a property you (or other tools) are interested in, yes, exactly. It gives them more information to work with, such as being able to calculate *how far* they are behind, while still carrying the information of whether they are behind at all. As for naming, `estimated_aired_episode` sounds good to me, but it should only be included for airing shows (and can be omitted otherwise).

RaitaroH: Just to confirm, this is your proposed layout, correct:
```json
{"title": "Bungou Stray Dogs 4th Season", "current_episode": "2", "total_episodes": "13", "score": "0", "status": "Ongoing", "estimated_aired_episode": 2}
{"title": "Flying Witch", "current_episode": "9", "total_episodes": "12", "score": "0", "status": "Finished"}
```
Perhaps the "status" is redundant if estimated exists, but I do think the value is useful.

If have tested on my end the new format, Let's say the previous status format looked nicer in awk, but either way works.

RaitaroH: If you think status shouldn't be printed at all, I can simply remove line 988.

FichteFoll: Works for me, but I'm also not sure what you as the consumer of this API need here. If what you have is sufficient, then I'm fine with either. I'm just making sure that what is being added is reasonably implemented.

trackma/ui/cli.py
FichteFoll: I'll leave this unresolved for @z411 to read.

trackma/ui/cli.py
RaitaroH: For me it works like this. I have slightly more info than `list` does.
I will update the description.

RaitaroH: @justchokingaround  and @anma-dev are the only people I know that may benefit from this json format.

RaitaroH: @FichteFoll I also added some code to print covers image urls, as my script adl can display images This way, it should be provider agnostic, as my previous implementation was using anilist only, which may cause issues on mal trackma output for example.

trackma/ui/cli.py
RaitaroH: Kind reminder.
@z411

Issue #624, opened 2022-07-03
Minimum supported Python version & formatting


FichteFoll:
The current code base is in a bit of a mess with its coding style of seemingly random line breaks to adhere to imho ridiculous code length requirements. Additionally, string formatting switches between % and str.format all the time and we might consider using the logging module instead of the custom messaging system as well.

As of today, the oldest Python version that has not reached [end of life](https://devguide.python.org/devcycle/#end-of-life-branches) is 3.7, which I strongly suggest to adopt as our minimum supported version, too. This will mean that we get access to some additions to the standard library as well as f-strings. There are tools to help with the migration to newer versions that I'd like to test & apply to the entire code base.

Regarding formatting, I have a personal preference on a couple of things but generally like to follow PEP8 with a focus on keeping line breaks clean from git diff collisions, i.e. always add trailing commas/punctuation where possible and binary operators after the line break. As for line length, 80 is ridiculous in my point of view. 100 is quite workable and 120 would be the maximum I'd allow for. 

Using an auto-formatter like `black` and enforcing it in some fashion (e.g. through a pre-commit hook) will probably go a long way here.

z411:
> The current code base is in a bit of a mess with its coding style of seemingly random line breaks to adhere to imho ridiculous line length requirements.

I'm not a fan of 80 char line requirements but they're mostly specified by PEP so it is what it is, I guess. But I don't think we should adhere to it so adamantly.

> we might consider using the logging module instead of the custom messaging system as well.

I thought of doing this long ago and I fully agree. The current messenger system is just a leftover because this program was originally written on Perl then ported over. Python's logging module is much cleaner and will allow for the current functionality and more (logging to file, logging to a window, etc.)

> his will mean that we get access to some additions to the standard library as well as f-strings and type hints.

I'm also a fan of the newer f-string style (I wasn't a fan of .format which is why I left a mess in there) so I'd be committed to switching to those fully. Type hints also look very clean to me and being a fan of typed languages in the first place I'd love having it implemented in the codebase.

> As for line length, 80 is ridiculous in my point of view. 100 is quite workable and 120 would be the maximum I'd allow for.

I agree. 100 is fine in my book.

> Using an auto-formatter like black and enforcing it in some fashion (e.g. through a pre-commit hook) will probably go a long way here. Ideally after most open pull requests have either been merged or rejected because it will cause conflicts with a lot of them.

[...]

PR #663, opened 2023-01-12
Create a wrapper around Anitopy as an alternative to AnimeInfoExtractor

FichteFoll:
I have been running a local branch configured with the anitopy parser of #591 for a while now and have also made some ad-hoc comparisons for the rescan results of my library. I have added a configuration to switch between the two parsers for everyone to try (but didn't add that to the GUIs so it needs to be edited in the config file directly). *AIE* stands for AnimeInfoExtractor.

### Comparisons

For context, all scans in this comparison where made against my local library which counts 1370 files (not all of which are anime or for shows on my list, but most are; so around ~1250).
Anitopy version used was [2.1.0.r2.fef4fc3](https://github.com/igorcmoura/anitopy/commit/fef4fc3f4817493b086561782e10bb333832a6d7). There have been a few bugfixes since then but nothing relevant for my library afaict.

|  | AIE | Anitopy |
|:-|:----|:--------|
|**Runtime** (`python -m trackma.ui.cli -d rescan`)|265s (1 run)|270s (3 runs)|

Why is this a table? Because I thought I'd have more things to compare easily, but the case-by-case comparisons make more sense in text.
That said, both were pretty much equal in what they matched properly (hint: almost everything, with proper altnames set up which wouldn't be the fault of the parser anyway, usually, except for some edge cases).

Talking about indvidual examples, here are some of my observations with anitopy (compared to AIE) and the library scans of above:

- Both strip underlines correctly if they are used consistently throughout the file name.
- Both recognize various episode formats (`S01E01`, `Episode 1`, `E01` `- 01`).
- Anitopy strips special characters at various positions, e.g. the dots in `Akuyaku Reijou ...` or the plus in `Cardfight!! Vanguard - will+Dress`. The former needs a different altname for AIE and Anitopy because of this. Appears to be configurable for Anitopy.
- Anitopy cannot handle `01v1.5` episodes numbers which AIE does. Whole numbers like `01v1` work, though. Shouldn't be hard to fix. Example: `Mahoutsukai Reimeiki E06v1.5 [1080p][AAC][JapDub][GerEngSub][Web-DL].mkv`
- Anitopy parses ``Kodocha - 001 I`m an Elementary School Student with an Agent.mkv`` to `Kodocha` which AIE does not.
- AIE more aggressively strips parentheses from names, which is worse for shows with a year e.g. `Urusei Yatsura (2022)`, but better for files with both a Japanese and English name, e.g. `[SlyFox] Summertime Rendering (Summer Time Render) - 12 [A746BE85].mkv` and `Yuuki Yuuna wa Yuusha-bu Shozoku (Yuki Yuna is a Hero - Club Member) - 01 [EngSub][BD 720p Hi444PP AAC].mkv`. Hard to determine in general whether parentheses should be stripped.
- AIE failed to parse `[The Impatient Miyafuji Kantai]  Luminous Witches - 06.5 [1080p][F48884A8].mkv`, converting it to `The Impatient Miyafuji Kantai`. Anitopy works. Should be an easy fix, though.

Not related to the parsers:

- `Yuuki Yuuna wa Yuusha de Aru - Hidamari - OVA` is mapped to `Yuki Yuna is a Hero: The Great Mankai Chapter - 1` (both produce this).

Packaging-related:

- AIE is a custom implementation while anitopy is a centralized and shared implementation based on Taiga's [anitomy](https://github.com/erengy/anitomy), where fixes to either repo should also be translated to the other. It even allows for updating the parsing library independent of trackma.
- AIE currently has tests inside the repo. Anitopy has its own tests and I would definitely be inclined to translate any test cases it's currently missing (at some point).
- Anitopy is a thirdy-party dependency. That means either depending on it optionally and providing AIE as a fallback or requiring it.

### Conclusions

In summary, I believe both work satisfactorily and can be used interchangeably. I don't want to pick a favorite here, so I'm proposing to add the *option* for Anitopy to be used if desired. We can always remove either later if we deem it useless.  In principle, I'm more inclined to Anitopy though because of it focusing on parsing file names only.

Furthermore, I also made some logging enhancements to make the following comparisons easier for me and found them to also be generally useful, e.g. when trying to diagnose mismatches in normal use.

I'm not too fond of the code architecture and quality of AIE, ~~but it's still much better than the Anitopy wrapper here. Especially the `hasattr` checks are atrocious. I don't feel like spending more time on this right (now) and~~ (fixed now) I have confirmed that it at least *works* and it also hasn't crashed on me once or similar. If a decision has been made to remove AIE, it'd be much easier to rewrite the wrapper to not have to imitate AIE's API.

Either way, I'd rather get this off my local branch than have it sit around for another 0.5 years.

Includes and closes #591.
Supersedes and closes #473.


z411:
The idea of having Anitopy as an option has been in my head for a while. Thanks, I'll test it myself. One thing; have you made performance comparisons? I'm a bit reluctant about adding additional parsing logic to the Anitopy parser but if it doesn't affect performance very much then it's ok. Scan list problems come more from our `difflib` usage; that should definitely need our attention.

FichteFoll:
Yes, I made performance comparisons (via `python -m trackma.ui.cli -d rescan`) and they are equally fast. I have included them in the original post but also edited a bit to clarify (just realized I forgot to add a unit). I expect most of the time to be spent matching against the list and not parsing the anime title, though.

I also figured that sleep isn't important for humans anyway and went ahead with refactoring to remote the `__dict__` abuse. It just didn't sit well. :sweat:

FichteFoll:
Since this is probably the biggest outstanding PR currently, I'd like to merge it soon-ish and continue with #624 when I find some free time.

PR #664, opened 2023-01-13
GTK: change "Play episode" menu aesthetic

imsamuka:
This continues the effort in #616, althrough, a simple one. 
I tested some designs in practice for some weeks, and ended with this, which is actually the same as the one i showed in https://github.com/z411/trackma/pull/616#discussion_r895049777.

## The PR:

It's simpler, because the "markings" now only convey a single piece of information: if the episode is available in the library or not; and, the information about being already watched or not, is still conveyed by the mere positioning of the numbers. This bring a more intuitive feel to the menu, as of before, a new user would probably take quite a while until they understand the meaning of the markings.


FichteFoll (approved):
This is definitely an improvement over the current status. Since I don't use the GTK UI, I won't put too much thought into how I would expect it (in the old PR I argued for inversing the condition for the "inconsistent" state, but this solution is easier to understand regardless).

PR #667, opened 2023-01-14
New syntax to CLI "play" command and "watch_continuously" config

imsamuka:
Most of what is implemented here, was discussed before on #484, which will be closed if accepted. If there's any disagreement here, please read the discussion there before.

On another note, my changes `Engine.play_episode()` don't break the `episode_missing` hook API: instead of passing a copy of the `ep_iter` iterator, only the first episode number is passed. But this severely limits the power of the API in the future. If permitted, i will break the API in favor of more flexibility with it (of course, updating the hook files in the repo as well).

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -932,7 +932,6 @@ class Engine:
             raise utils.EngineError('Show given is invalid')
 
         if isinstance(playep, int) or isinstance(playep, str):
-            ep_iter = ()
             try:
                 first_ep = int(playep)
             except ValueError:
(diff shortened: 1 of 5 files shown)


imsamuka:
Since no more input was given, I did some changes to conform to the avoidance of breaking changes.

* Absolute episodes are no longer prepended by `#`.
* Relative episode numbers are now prepended by `n`, standing for _next_.
* The empty `""` range behavior is now dependent on `watch_continuously`.
* `watch_continuously` is now `False` by default, maintaining the original behavior.
* Improved the help message for `play` (debatable, but i think it's pretty intelligible)

PR #672, opened 2023-02-05
Show time left for episodes on release

JafarAbbas33:
This pull request shows the time left for the next episode which is going to be released (if any) and removes info cache when asked to download data.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+Pillow
+PyGobject
--- a/trackma/data.py
+++ b/trackma/data.py
@@ -544,8 +544,13 @@ class Data:
         utils.save_data(self.meta, self.meta_file)
 
     def download_data(self):
-        """Downloads the remote list and overwrites the cache"""
+        """Downloads the remote list and overwrites the cache and info"""
         self.showlist = self.api.fetch_list()
+        try:
+            os.remove(self.info_file)
+        except Exception:
+            pass
+        self.infocache.clear()
 
         if self.api.api_info['merge']:
             # The API needs information to be merged from the
--- a/trackma/lib/libanilist.py
+++ b/trackma/lib/libanilist.py
@@ -483,13 +483,13 @@ fragment mediaListEntry on MediaList {
                 ('Average score',   item.get('averageScore')),
                 ('Status',          self._translate_status(item['status'])),
             ]
-        })
+        }
         if item.get('nextAiringEpisode', None):
             next_episode_release_time = self._int2date(item.get('nextAiringEpisode').get('airingAt'))
             time_diff  = next_episode_release_time - datetime.datetime.now()
             show_info['extra'].append(('Next Episode', item.get('nextAiringEpisode').get('episode')))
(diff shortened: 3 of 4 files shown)


trackma/utils.py
```diff
@@ -229,6 +230,16 @@ def load_data(filename):
 
 
 def save_data(data, filename):
+    if '.list' in filename:
+        path = expand_path('~/Desktop/Coding_Projects/new-os-setup/linux-backup/anilist.backup')
```
FichteFoll, line 234: This hard-coded past needs to be replaced with a dynamic one if this is meant to stay. Not really sure what the purpose of this JSON backup is, though. Sounds like a better idea for this would be to move the old file to a backup path first and then start writing the main file.

trackma/lib/libanilist.py
```diff
@@ -482,7 +483,14 @@ def _parse_info(self, item):
                 ('Average score',   item.get('averageScore')),
                 ('Status',          self._translate_status(item['status'])),
             ]
-        })
+        }
+        if item.get('nextAiringEpisode', None):
```
FichteFoll, line 487:
```suggestion
        if item.get('nextAiringEpisode'):
```

trackma/lib/libanilist.py
```diff
@@ -482,7 +483,14 @@ def _parse_info(self, item):
...
+            next_episode_release_time = self._int2date(item.get('nextAiringEpisode').get('airingAt'))
+            time_diff  = next_episode_release_time - datetime.datetime.now()
+            show_info['extra'].append(('Next Episode', item.get('nextAiringEpisode').get('episode')))
+            show_info['extra'].append(('Next Episode In', str(time_diff)))
+        show_info['extra'].append(('Information fetched on', datetime.datetime.now().strftime("%d %B, %Y at %I:%M %p")))
```
FichteFoll, line 492: From your testing or knowledge, how ofting is this information refetched? It doesn't really help if this needs to be requeried manually. Perhaps it should even be invalidated/ignored if the timestamp is in the past already.

trackma/data.py
```diff
@@ -544,8 +544,13 @@ def _save_meta(self):
         utils.save_data(self.meta, self.meta_file)
 
     def download_data(self):
-        """Downloads the remote list and overwrites the cache"""
+        """Downloads the remote list and overwrites the cache and info"""
         self.showlist = self.api.fetch_list()
+        try:
+            os.remove(self.info_file)
```
FichteFoll, line 550: What issue does deleting the file first solve?

FichteFoll (changes requested):

requirements.txt
```diff
@@ -0,0 +1,2 @@
+Pillow
+PyGobject
```
FichteFoll, line 2: Please remove this incomplete list of dependencies.

trackma/lib/libanilist.py
Birdulon: Probably makes more sense to ensure that UIs just show the information in the info view, rather than hack it onto the `extra` field on a per-API basis.

Issue #698, opened 2023-07-04
Poster image doesn't load -- module 'PIL.Image' has no attribute 'ANTIALIAS'


uzuto:
Error Log for trackma-gtk

```
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/site-packages/trackma/ui/gtk/imagebox.py", line 44, in run
    self._save_image(self._download_file())
  File "/usr/lib/python3.11/site-packages/trackma/ui/gtk/imagebox.py", line 61, in _save_image
    image.thumbnail((self._width, self._height), Image.ANTIALIAS)
                                                 ^^^^^^^^^^^^^^^
AttributeError: module 'PIL.Image' has no attribute 'ANTIALIAS'
```

Error Log for trackma-qt

```
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/trackma/ui/qt/workers.py", line 58, in run
    im.thumbnail((self.size[0], self.size[1]), Image.ANTIALIAS)
                                               ^^^^^^^^^^^^^^^
AttributeError: module 'PIL.Image' has no attribute 'ANTIALIAS'
[1]    12059 IOT instruction (core dumped)  trackma-qt
```

I have python-pillow installed.
OS : Arch Linux

molkoback:
Fix here: https://github.com/z411/trackma/pull/697

uzuto:
Thanks! replacing image.ANTIALIAS with image.LANCZOS worked but its very slow

z411:
Is it slower? I thought Antialias was an alias for Lanczos. Although now that I think of it Lanczos is probably overkill.

molkoback:
They are the same.
`>>> Image.ANTIALIAS == Image.LANCZOS
True`

PR #700, opened 2023-07-11
Replace Image.ANTIALIAS with Image.BICUBIC in QT and GTK

molkoback:
Image.ANTIALIAS has been removed in PIL 10.0.0: https://pillow.readthedocs.io/en/stable/deprecations.html
Image.LANCZOS is identical but [it seems like overkill](https://github.com/z411/trackma/issues/698#issuecomment-1627437824).


FichteFoll (approved):

FichteFoll:
I agree that bicubic should be fine here.

Issue #655, opened 2022-11-08
Relations/Franchise/Related


MorsMortium:
Each of the 5 websites has a way to display related content to the currently viewed. It would be nice, if these could be listed and opened, in the program

toprak:
Yes it would be awesome! also adding staff would be good.

PR #705, opened 2023-08-23
add related anime section to gtk interface when using mal api

MoreOneNight6:
2023-08-23-193119_644x837_scrot


MoreOneNight6:
#655

PR #722, opened 2024-01-01
New 'Next episode column' in GTK and for MAL (Qt and GTK) + miscellaneous code refactor

BigBoyBarney:
I now realise this probably should have been a few different PRs instead of one.

# Changelist
Screenshots are at the bottom
### UI
- Implemented 'Next episode' column with custom sorting
- QoL changes to right-click context menu:
	- Play Next	→ Play next episode
	- Show details... 	→ Show details
	- Open web site	→ Open on [website], where [website] is the currently used API. Tested with MAL and Anilist.
	- Set alternate title...	→ Set alternate title
- Sorting indicators no longer take up screen space on unsorted columns
- Centred columns other than 'Title'
- Increased default window size to 1080x720 (from 740x480). The current default was set in 2015, and it is too small for modern displays.
- Renamed some column titles (see screenshots). Default visible columns had to be reset in order to accommodate this change:
	- Progress	→ Watched
	- Percent → Progress
- Changed default sorting directions for some columns to be more intuitive, defaulting to descending order on the first click. 
E.g: 10, 10, 8, 4 etc. Changed columns:
	- Watched
	- Score
	- Next episode
	- Progress

### Code
- Implemented 'Next episode' column with custom sorting
- New function to reset visible columns in the config, to ensure the default columns are visible even after being renamed. A more elegant solution would be to completely rewrite how visible columns are handled and *not* reference the column titles directly.
- Rewrote parts of `ShowTreeView.py`:
	- Row data access and general column referencing is done through a `TreeConstants` class to improve readability and uniformity
	- TreeView model is now set immediately after initialisation
	- Completely rewrote how sorting is set up

### UI
- Increased default window size to 1080x720 (from 740x480)
-  Implemented 'Next episode' column for MAL (Anilist was already supported)

### Code
- Replaced deprecated `datetime.utcnow()` with time zone aware `datetime.now(datetime.timezone.utc)`
- Implemented 'Next episode' column for MAL. The code to display the datetime object was already written by someone else — most likely Z — and I did not touch that too much, which is why the way Qt displays the time differs from GTK.

## Other
- Minimum python version bumped to 3.10, as it introduces a lot of QoL changes. Currently one of the oldest "major" distros, Ubuntu 22.04, ships with Python 3.10 already, so pretty much every other distro should be at least this or newer.
- Updated `poetry.lock`

# Screenshots GTK
Default
Default

Next episode sorting down (Anilist)
Next episode sorting down

Next episode sorting up (Anilist)
Next episode sorting up

MAL Support
image

Context menu Anilist
Context menu Anilist

Context menu MAL
Context menu MAL

# Screenshots Qt
Anilist
image

MAL
image


BigBoyBarney:
Added MAL support, with all of their information. Kitsu provides no information and Shikimori is dead. 
'Next episode' functionality can be considered complete at this point IMO.
Updated the OP with Qt images.

The poetry update consists mostly of minor version changes, the only differences being [urllib3](https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html), but I don't think this is an issue for Trackma, as only barebones urllib functionality is being used.

The other major change is obviously python 3.8 → 3.10, but I think this shouldn't be an issue either.

---

Could someone take a look? I promise the code is readable and well commented, so it shouldn't take too long. The GTK version is well tested, but Qt should be stable as well.
@z411

v-fox:
Doesn't apply cleanly to master due to some extraneous changes but easy to clean up by hand. Absolutely essential feature, especially now that Qt5 is obsoleted but PyQt6 port is stalled.
But I wish it would show hours too.

FichteFoll:
Thank you for the detailed summary in the PR description. These definitely sound like useful changes and are definitely something I'd like to review soonish. However, I spent my time today on other smaller and arguably more important issues, so this will have to wait a bit longer. Just know that I/we intend to get this in eventually.

Just one thing to note: Imo the default column order should be indentical between gtk and qt. I did not check whether that is the case in code (since you can re-order columns in the Qt UI) but if this isn't already the case, it will definitely something I would request or adjust myself.

BigBoyBarney:
Thanks for taking a look!

I since stopped using Trackma and have no particular desire to work on this anymore, so feel free to merge, squash or otherwise delete anything you'd like!

PR #727, opened 2024-01-12
Unification of 'open-folder' function, should fix 'xdg-open' path error

### LINKED ISSUE #726: "Open containing folder" action uses a hard-coded path for `xdg-open`
The hard-coded path `/usr/bin/xdg-open` may not exist (such as when using NixOS), causing the "open containing folder" action to silently fail. This should simply use the appropriate `$PATH` entry (as with the macOS `open` added in https://github.com/z411/trackma/pull/447).

Logs from a failure:
```
  File "/nix/store/d37h5x02hg7d43nb8rdzwf76jyz05xbn-trackma-0.8.6/lib/python3.11/site-packages/trackma/ui/gtk/window.py", line 589, in _open_folder
    subprocess.Popen(["/usr/bin/xdg-open", os.path.dirname(filename)],
  File "/nix/store/5k91mg4qjylxbfvrv748smfh51ppjq0g-python3-3.11.6/lib/python3.11/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/nix/store/5k91mg4qjylxbfvrv748smfh51ppjq0g-python3-3.11.6/lib/python3.11/subprocess.py", line 1950, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/bin/xdg-open'

During handling of the above exception, another exception occurred:

  File "/nix/store/d37h5x02hg7d43nb8rdzwf76jyz05xbn-trackma-0.8.6/lib/python3.11/site-packages/trackma/ui/gtk/window.py", line 504, in _on_show_action
    self._open_folder(*data)
  File "/nix/store/d37h5x02hg7d43nb8rdzwf76jyz05xbn-trackma-0.8.6/lib/python3.11/site-packages/trackma/ui/gtk/window.py", line 594, in _open_folder
    raise utils.EngineError("Could not open folder.")
trackma.utils.EngineError: Could not open folder.
```

BigBoyBarney:
Should fix #726 as long as `xdg-open` is in `$PATH`, which can be assumed for every distro. I tested it for GTK, Qt and CLI on Fedora 39.

> [!CAUTION]
> **As my other PR (#722), this requires python 3.10. Do not merge before #722 or without updating python dep.**

# Changes
## Code
- Unified all 'open folder' functions in `utils.py`. Can take an `error_callback` param so error reporting can still be done uniquely for every UI.
- Fixed OSError not being raised when the folder was in the library, but was changed after scanning. I.e:
	1. Start trackma
	2. 'Show X' gets scanned and found
	3. Rename folder of 'show X'
	4. Open folder
	5. Error is raised correctly

### DIFF
--- a/trackma/utils.py
+++ b/trackma/utils.py
@@ -490,6 +490,37 @@ def get_terminal_size(fd=1):
     return hw
 
 
+def open_folder(engine, show_id, error_callback=None):
+    try:
+        show_to_open = engine.get_show_info(show_id)
+        filename = engine.get_episode_path(show_to_open)
+        with open(os.devnull, 'wb') as DEVNULL:
+            if sys.platform == 'darwin':
+                subprocess.Popen(["open", os.path.dirname(filename)],
+                                 stdout=DEVNULL,
+                                 stderr=DEVNULL)
+            elif sys.platform == 'win32':
+                subprocess.Popen(["explorer", os.path.dirname(filename)],
+                                 stdout=DEVNULL,
+                                 stderr=DEVNULL)
+            else:
+                subprocess.Popen(["xdg-open", os.path.dirname(filename)],
+                                 stdout=DEVNULL,
+                                 stderr=DEVNULL)
+    except OSError:
+        # Open failed.
+        if error_callback:
+            error_callback("Could not open folder.")
+        else:
+            raise EngineError("Could not open folder.")
+    except EngineError:
+        # Show not in library.
+        if error_callback:
+            error_callback("No folder found.")
+        else:
+            raise EngineError("No folder found.")
+
+
 def show():
     return {
         'id':           0,
--- a/trackma/utils.py
+++ b/trackma/utils.py
@@ -507,18 +507,11 @@ def open_folder(engine, show_id, error_callback=None):
                 subprocess.Popen(["xdg-open", os.path.dirname(filename)],
                                  stdout=DEVNULL,
                                  stderr=DEVNULL)
-    except OSError:
-        # Open failed.
+    except Exception as e:
         if error_callback:
-            error_callback("Could not open folder.")
+            error_callback(e)
         else:
-            raise EngineError("Could not open folder.")
-    except EngineError:
-        # Show not in library.
-        if error_callback:
-            error_callback("No folder found.")
-        else:
-            raise EngineError("No folder found.")
+            raise e
 
 
 def show():
--- a/trackma/utils.py
+++ b/trackma/utils.py
@@ -494,24 +494,35 @@ def open_folder(engine, show_id, error_callback=None):
     try:
         show_to_open = engine.get_show_info(show_id)
         filename = engine.get_episode_path(show_to_open)
+
         with open(os.devnull, 'wb') as DEVNULL:
-            if sys.platform == 'darwin':
-                subprocess.Popen(["open", os.path.dirname(filename)],
-                                 stdout=DEVNULL,
-                                 stderr=DEVNULL)
-            elif sys.platform == 'win32':
-                subprocess.Popen(["explorer", os.path.dirname(filename)],
-                                 stdout=DEVNULL,
-                                 stderr=DEVNULL)
-            else:
-                subprocess.Popen(["xdg-open", os.path.dirname(filename)],
-                                 stdout=DEVNULL,
-                                 stderr=DEVNULL)
+            command = []
+
+            match sys.platform:
+                case 'darwin':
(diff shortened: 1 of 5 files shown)


trackma/utils.py
```diff
@@ -490,6 +490,53 @@ def get_terminal_size(fd=1):
...
+        if error_callback:
+            error_callback(f'Could not open folder.\n{e}')
+        else:
+            raise e
+
+    except Exception as e:
+        if error_callback:
+            error_callback(e)
+        else:
+            raise e
```
FichteFoll, line 537: This is an anti-pattern. Either the `open_folder` function should handle errors on its own, wrap errors with custom errors or let them bubble, but making this depend on a callback parameter is an unnecessary level of indirection.

With respect to the function providing *some* custom handling for annotating `OSError`s with "Could not open folder", I suggest to wrap only `OSError`s with a custom exception and otherwise let the callers handle the exceptions with their respective UI options.

FichteFoll (changes requested):
Thanks for the changes.

Which part of this PR requires Python 3.10? I'm not seeing it …

BigBoyBarney:
> Thanks for the changes.
> 
> Which part of this PR requires Python 3.10? I'm not seeing it …

```python
match sys.platform:
            case 'darwin':
                command = ["open", os.path.dirname(filename)]
            case 'win32':
                command = ["explorer", os.path.dirname(filename)]
            case _:
                command = ["xdg-open", os.path.dirname(filename)]
                ```
Match was introduced in Python 3.10

File: trackma/utils.py (hunk 1 above)
Comment (FichteFoll) [reply trackma/utils.py:537]: `try: open_foolder(…); catch: …`

Comment (BigBoyBarney) [reply trackma/utils.py:537]: Regarding the OSError message.
image
I added the string for clarity, but it's not really necessary if you're opposed to it.

If you want to do it through `try: catch:` then you'd have to add that everywhere, where you call open_folder(). Since the function is super simple and its usage doesn't vary in the slightest, using the callback function saves time. I can remove if it and change the wrap the calls in `try:` if you prefer that though

Comment (FichteFoll) [reply trackma/utils.py:537]: Wrapping the error is a valid use case (that I also mentioned in my initial comment). It shouldn't depend on whether an error callback was provided, however, since that's what we have try-catch for. Please just remove the callback parameter and handle thown exceptions at the call site.

(In general, I prefer error handling like it is provided by Rust (or even Golang to some extent), but Python has exceptions and we should use them as they are intended since the rest of the ecosystem is built around them.)

Comment (BigBoyBarney) [reply trackma/utils.py:537]: Just to make sure I understand what you mean.

Like this?
in `utils.py`
```python
def open_folder(engine, show_id):
    """
    Open the folder containing the show of the passed ID.
    """

    try:
        show_to_open = engine.get_show_info(show_id)
        filename = engine.get_episode_path(show_to_open)

        with open(os.devnull, 'wb') as DEVNULL:
            command = []

            match sys.platform:
                case 'darwin':
                    command = ["open", os.path.dirname(filename)]
                case 'win32':
                    command = ["explorer", os.path.dirname(filename)]
                case _:
                    command = ["xdg-open", os.path.dirname(filename)]

            process = subprocess.Popen(command,
                                       stdout=DEVNULL,
                                       stderr=subprocess.PIPE,
                                       universal_newlines=True)

            _, stderr = process.communicate()

            if process.returncode != 0:
                # Command failed.
                raise OSError(stderr)

    except OSError as e:
        raise OSError(f'Could not open folder: {e}')

    except Exception as e:
        raise RuntimeError(f'An error occurred while trying to open the folder: {e}')
```


in (gtk) `window.py`
```python
    def _open_folder(self, show_id):
        try:
            utils.open_folder(self._engine, show_id)
        except Exception as e:
            self._error_dialog_idle(e)
```

Or do you want me to create a new
```python
class FolderError(TrackmaError):
    pass
```

FichteFoll: Yes, that is basically what I mean. Either would be fine imo, but a custom `OpenFolderError(TrackmaError)` would be most appropriate since we're interpreting the underlying errors already when re-throwing a wrapped exception. Additionally, you could directly throw a `OpenFolderError` when the subprocess's return code is 0.

z411: Alternatively you could just raise an EngineError and catch it outside, which is what all UIs do right now.

FichteFoll:
I implemented this as I described in my comments in #770, which supersedes this PR. Thanks for getting this started either way.

PR #746, opened 2025-01-05
Hooks: Use `load_hook` to support python >= 3.12 | < 3.5

### LINKED ISSUE #736: 'FileFinder' object has no attribute 'find_module'
The engine fails to initialize with the error `AttributeError: 'FileFinder' object has no attribute 'find_module'`. It happens because it was deprecated, see https://docs.python.org/3/library/pkgutil.html#pkgutil.find_loader.

It can be solved easily changing this:

https://github.com/z411/trackma/blob/e56718b3ebcadc1a3ce73d6530d1c342f8ab80ed/trackma/engine.py#L315

For this:

```python
module = loader.find_spec(name).loader.load_module(name)
```

I think maybe there's a better way to do it(?) but I don't have much time to check it.

ahmubashshir:
Fixes #736
Signed-off-by: Mubashshir <ahmubashshir@gmail.com>

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -304,6 +304,18 @@ class Engine:
             if os.path.isdir(hooks_dir):
                 import pkgutil
 
+                try:
+                    from importlib.util import module_from_spec
+
+                    def load_hook(loader, name):
+                        spec = loader.find_spec(name)
+                        module = module_from_spec(spec)
+                        spec.loader.exec_module(module)
+                        return module
+                except ImportError:
+                    def load_hook(loader, name):
+                        return loader.find_module(name).load_module(name)
+
                 self.msg.info("Importing user hooks...")
                 for loader, name, ispkg in pkgutil.iter_modules([hooks_dir]):
                     # List all the hook files in the hooks folder, import them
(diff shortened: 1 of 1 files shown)


trackma/engine.py
```diff
@@ -304,6 +304,18 @@ def start(self):
...
+                    from importlib.util import module_from_spec
+
+                    def load_hook(loader, name):
+                        spec = loader.find_spec(name)
+                        module = module_from_spec(spec)
+                        spec.loader.exec_module(module)
+                        return module
+                except ImportError:
+                    def load_hook(loader, name):
+                        return loader.find_module(name).load_module(name)
```
FichteFoll, line 317: trackma currently requires Python 3.8, so we don't need this fallback code for <3.5 (which is ancient).

FichteFoll (changes requested):

trackma/engine.py
ahmubashshir: I'm stuck in implementing this here... `loader.find_module().load_module()` allowed importing sibling packages, but newer one doesn't... so I'm stuck on this... Is there any way to do this without resorting to sys.path modification?

FichteFoll: Guess I'll have to take a look at testing this myself. No promises when I'll get to that, though.

ahmubashshir: Found a way to fix sibling imports, but it's not clean...

I should move load_module to utils module...

FichteFoll:
After some research I am confident that #769 is the proper way to do this now, so I'm closing this PR as a duplicate. Thanks for taking a look.

PR #772, opened 2025-04-23
Module for loading hooks

ahmubashshir:
- **hooks: Implement hook loader module**
- **engine: Use trackma.hooks to load hooks**

This approach seems cleaner to me than #769, and also has the added benefit of not breaking unless python ignores `__path__`

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -31,6 +31,7 @@ from trackma import messenger
 from trackma import utils
 from trackma.extras import redirections
 from trackma.parser import get_parser_class
+from trackma import hooks
 
 
 class Engine:
@@ -302,22 +303,12 @@ class Engine:
 
         # Load hook files
         if self.config['use_hooks']:
-            hooks_dir = utils.to_config_path('hooks')
-            if os.path.isdir(hooks_dir):
-                import importlib.util
-                import pkgutil
-
+            if hooks.init():
                 self.msg.info("Importing user hooks...")
-                for finder, name, _ in pkgutil.iter_modules([hooks_dir]):
-                    # List all the hook files in the hooks folder, import them
-                    # and call the init() function if they have them
-                    # We build the list "hooks available" with the loaded modules
-                    # for later calls.
+                for name in hooks.iter():
                     try:
                         self.msg.debug("Importing hook {}...".format(name))
-                        module_spec = finder.find_spec(name)
-                        module = importlib.util.module_from_spec(module_spec)
-                        module_spec.loader.exec_module(module)
+                        module = hooks.load(name)
                         if hasattr(module, 'init'):
                             module.init(self)
                         self.hooks_available.append(module)
--- a/trackma/hooks.py
+++ b/trackma/hooks.py
@@ -1,0 +1,33 @@
+__path__ = []
+
+
+def iter():
+    """Return an iterator over the modules in __path__."""
+    if not __path__:
+        return ()
+
+    from pkgutil import iter_modules
+    return (name for _, name, _ in iter_modules(__path__))
+
+
+def load(name, attr=None):
+    if not __path__:
+        raise ImportError(f"Cannot load hook '{name}'")
+
+    from importlib import import_module
+    return import_module(f'trackma.hooks.{name}')
+
+
+def init():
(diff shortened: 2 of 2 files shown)


molkoback:
This doesn't seem right. The current implementation loads hooks directly from ~/.trackma/hooks while your implementation checks which hooks are in that directory and loads them from trackma.hooks.{name}. So if I build a custom hook I have to put it in two places.

The idea of creating a separate module for hooks is good though.

molkoback:
Wait, how does this work? At which point do you make the found hooks members trackma.hooks so you can import them using `import_module(f'trackma.hooks.{name}')`?

The idea of making hooks their separate module is good though. It should probably be a class to avoid __path__.

ahmubashshir:
> Wait, how does this work? At which point do you make the found hooks members trackma.hooks so you can import them using `import_module(f'trackma.hooks.{name}')`?

the `__path__` is set after `hooks.init()` is called, so to use `import_module(...)` for `trackma.hooks.*` you'd have to call that.

> The idea of making hooks their separate module is good though. It should probably be a class to avoid `__path__`.

Everything is done by python... I implemented it using sys.meta_path first, but just setting `__path__` seemed cleaner to me, and it's also documented by python[^1]...


[^1]: https://docs.python.org/3/tutorial/modules.html#packages-in-multiple-directories

trackma/engine.py
```diff
@@ -302,22 +303,11 @@ def start(self):
...
+            if hooks.init():
                 self.msg.info("Importing user hooks...")
-                for finder, name, _ in pkgutil.iter_modules([hooks_dir]):
-                    # List all the hook files in the hooks folder, import them
-                    # and call the init() function if they have them
-                    # We build the list "hooks available" with the loaded modules
-                    # for later calls.
+                for name in hooks.iter():
                     try:
-                        self.msg.debug("Importing hook {}...".format(name))
```
FichteFoll, line 317: This debug message seems too useful to be removed.

trackma/hooks.py
```diff
@@ -0,0 +1,33 @@
...
+    from pkgutil import iter_modules
+    return (name for _, name, _ in iter_modules(__path__))
+
+
+def load(name, attr=None):
+    if not __path__:
+        raise ImportError(f"Cannot load hook '{name}'")
+
+    from importlib import import_module
+    return import_module(f'trackma.hooks.{name}')
```
FichteFoll, line 18: Using `iter_modules` to iterate modules in a folder, only to throw away all but one part of the return values, still seems like a waste to me, but I also realize that using `import_module` and the previous way of managing hooks has different (side) effects. Notably I assume that the modules were not put into `sys.modules` before, which may or may not be desirable. Building a module path string can also be fragile in the sense that `iter_modules` might return module names that cannot be imported like this (e.g. when they contain hyphens or spaces), but I don't actually know about this. Users should not use such characters in their hook file names either way.

FichteFoll (commented):
The idea to use `__path__` to magically group all hooks into a `trackma.hooks` package sounds neat and offloading some of the hook logic into a separate file is also appreciated since `engine.py` definitely does too much currently.

However, I don't feel confident enough to make the shot regarding using `import_module` here.

(This means it's not a blocker, but I also don't want to call the shot to merge it.)

trackma/hooks.py
ahmubashshir: We can urlencode the name and replace `%` with `_`, or throw an error if hook name contains invalid characters?

I'd prefer this approach (turning `trackma.hooks` into a namespace package), cause now I can share code between hooks (I have a `auth` hook, that only provides `init/destroy/get_creds`, and keeps credentials in libsecret, that is used by other hooks)

ahmubashshir: nvm, python doesn't enforce valid module name if using `import_module`, only disallowed character seems to be `.`

```
[D] Engine: Importing hook local test...
<module 'trackma.hooks.local test' from '/home/ahm/.config/trackma/hooks/local test.py'>
[D] Engine: Importing hook local-test...
<module 'trackma.hooks.local-test' from '/home/ahm/.config/trackma/hooks/local-test.py'>
```

PR #773, opened 2025-04-28
replace Shikimori domain to avoid redirection problems

quardbreak:
Just noticed it uses old domain which can't be opened in browser (ERR_TOO_MANY_REDIRECTS). I believe adjusting domain to the current shikimori.one is fine and shouldn't cause any troubles.


FichteFoll (approved):
Thanks. From what I can tell after some search engine consultation, this site does look like a legitimate replacement, though I have never visited either of these domains before.

PR #784, opened 2025-11-29
Promote "ONA" to a first-class citizen anime type

### LINKED ISSUE #781: Show ONA as ONA instead of OVA
Hello

When searching for Sakamoto Days, we receive using the kitsu endpoint 4 results:

- Nichijou as TV
- Sakamoto Days as OVA
- Sakamoto Days Part 2 as OVA
- and a movie

The kitsu endpoint displays them as ONA (`..."subtype": "ONA"...`)

Is it possible to show it as ONA on the searching list?

From my research, it seems related to this line:https://github.com/z411/trackma/blob/062faf62c315884a1b6885d63d4e13f12eb6af44/trackma/utils.py#L131

Thanks

FichteFoll:
Closes #781

~~Note that I did not test this in the QT UI. I probably should …~~

### DIFF
--- a/trackma/ui/qt/models.py
+++ b/trackma/ui/qt/models.py
@@ -353,11 +353,11 @@ class AddListModel(QtCore.QAbstractListModel):
             return self.thumbs.get(row)
         elif role == QtCore.Qt.BackgroundRole:
             t = self.results[row].get('type')
-            if t == utils.Type.TV or t == utils.Type.ONA:
+            if t == utils.Type.TV:
                 return QtGui.QColor(202, 253, 150)
             elif t == utils.Type.MOVIE:
                 return QtGui.QColor(150, 202, 253)
-            elif t == utils.Type.OVA:
+            elif t == utils.Type.OVA or t == utils.Type.ONA:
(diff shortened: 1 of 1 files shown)


trackma/ui/qt/models.py
```diff
@@ -360,7 +353,7 @@ def data(self, index, role):
             return self.thumbs.get(row)
         elif role == QtCore.Qt.BackgroundRole:
             t = self.results[row].get('type')
-            if t == utils.Type.TV:
+            if t == utils.Type.TV or t == utils.Type.ONA:
```
z411: As I mentioned in #781 I feel like ONAs are more similar than OVAs than TV shows - could we use the OVA color instead? Or, maybe a new one?

FichteFoll:
Note that this implementation does not migrate already cached data or force it to be refreshed.

z411:
You can force a refresh by bumping the database version number: https://github.com/z411/trackma/blob/8cacb7311f2dadd3986e98faefe51a2e30f5a9b3/trackma/data.py#L39

PR #801, opened 2025-12-19
Update view during queue processing

### LINKED ISSUE #800: App crash on autoretrieve with pending queue
This applies to the Qt interface only (v0.9).

During app start, if the queue has pending items, autoretrieve would crash the app.

Steps to reproduce:

- Set Settings > Sync > Autoretrieve to "Always at start"

- Uncheck Settings > Sync > Autosend > "at exit"

Apply those settings and return to the main window. Now add items to the pending queue and restart the app. It will crash with this error:

```
Showing debug information.
Engine: Trackma v0.9 - using account redacted(mal).
Data: Initializing...
[D] Data: Reading userconfig...
libmal: Initializing...
Data: Using libmal (anime)
[D] Engine: Starting engine...
[D] Data: Locking database...
[D] Data: Reading metadata...
[D] Data: Reading queue...
[D] Data: Reading info DB...
Data: Processing queue...
[D] Data: Reading cache...
libmal: Updating item Redacted...
[D] libmal: POST data: b'status=on_hold'
[D] libmal: PATCH URL: https://api.myanimelist.net/v2/anime/0000/my_list_status
  File "/redacted/python3.13/site-packages/trackma/ui/qt/mainwindow.py", line 1309, in ws_changed_show
    self.view.model().sourceModel().update(show['id'], is_playing)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/redacted/python3.13/site-packages/trackma/ui/qt/models.py", line 130, in update
    row = self.id_map[showid]
          ^^^^^^^^^^^
AttributeError: 'ShowListModel' object has no attribute 'id_map'
```

This does not happen when manually retrieving the lists from the menu, only when the app is starting.

I will open a pr that tries to fix this.

m0uss3:
This fixes a crash when the autoretrieve logic tries to deal with pending items in the queue. The engine calls the start method of the Data class during its own initialization:

https://github.com/z411/trackma/blob/v0.9/trackma/engine.py#L253

and tries to process the queued items before retrieving the lists from the remote:

https://github.com/z411/trackma/blob/v0.9/trackma/data.py#L168

this would trigger the `ws_changed_show` method of `MainWindow` which updates the view and expects the model to be  available but it isn't because the engine has not finished its own initialization yet (the model would be initialized when the engine has finished starting here I think https://github.com/z411/trackma/blob/v0.9/trackma/ui/qt/mainwindow.py#L1417). Here it calls a model proxy I think:

https://github.com/z411/trackma/blob/v0.9/trackma/ui/qt/mainwindow.py#L1304

and crashes because the model attribute `id_map` is not set:

https://github.com/z411/trackma/blob/v0.9/trackma/ui/qt/models.py#L130

The fix in this pr is simple and it "just works" but I wonder if there's a better one? Would it be better to refactor the queue processing logic out of the Data.start method? or make the queue wait for the engine to start?

Fixes #800

### DIFF
--- a/trackma/ui/qt/mainwindow.py
+++ b/trackma/ui/qt/mainwindow.py
@@ -772,11 +772,9 @@ class MainWindow(QMainWindow):
 
         if pyqt_version == 5:
             self.view.horizontalHeader().setSectionResizeMode(1, QHeaderView.Stretch)
-            self.view.horizontalHeader().setSectionResizeMode(2, QHeaderView.Fixed)
             self.view.horizontalHeader().setSectionResizeMode(3, QHeaderView.Fixed)
         else:
             self.view.horizontalHeader().setResizeMode(1, QHeaderView.Stretch)
-            self.view.horizontalHeader().setResizeMode(2, QHeaderView.Fixed)
             self.view.horizontalHeader().setResizeMode(3, QHeaderView.Fixed)
 
         # Recover column state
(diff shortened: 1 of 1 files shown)


m0uss3:
Closed as already fixed in master

PR #807, opened 2025-12-21
Add last updated column

### LINKED ISSUE #732: [Request] "Last Updated" Column
I'll typically watch several shows with different groups of people, which keeps my Currently Watching tab rather busy. An optional column that includes the last date a shows status was changed (watched an episode, moved from Oh Hold to Currently Watching) would be a big help for keeping the freshest shows at the top of the list.

### LINKED ISSUE #799: [feature request] please add "last updated" column
hello,

Could you add "Last Updated" column to the anime/manga list please. it basically shows a time when the entry was updated and its a best way to sort through the list. Taiga has the same feature as well.

m0uss3:
Added the multiple times requested "last updated" column, although it was named "My update" to match the other "My start" and "My finish" columns in style. Arranged between the start and finish columns in the column selector to follow a logical order. 

- Implemented for Qt and GTK GUIs for Anilist, MAL, and Kitsu API providers.
- Not implemented for Shikimori due to the API server being down, and VNDB, since their API does not seem to track granular updates over time.

Note that cached lists would show an empty value for "My update" column until the user manually resyncs or redownloads the list, but is compatible with previously cached lists.
Note that the update date value changes at least when the status, score, and progress change.

Closes #732
Closes #799


m0uss3:
Current progress:

update-min

The column is sortable, uses timestamp for comparisons under the hood. Backwards compatible with previous cached data. Implements a timer to refresh the relative date column in real time every 10 seconds. Dates are stored in timezone aware utc datetime objects. 

The tooltip displays the date/time in the user's local format:

tooltip-min

Users will need to resync their lists to see the update date and until then the column displays a default value "-" with informative tooltip.

The new column is available in the header context menu:

header-min

m0uss3:
Current progress (GTK frontend):

Column:

col-min

Tooltip:

tip-min

Context menu:

menu-min

z411:
Great work so far. Before you continue I'd like to ask:

I reckon this takes the last updated time from the item in the remote. APIs update this timestamp whenever we upload our changes. If I set my list to send changes every 3 hours, but I watch 5 episodes in that timeframe. Wouldn't the last updated timestamp be the time whenever these changes got uploaded? Wouldn't this column be closer to "Last synced" instead?

If that's the case, I wonder if we could keep a separate local database with the timestamps of the local changes, and merge them with the timestamps of the remote sync times provided by the remote whenever we sync the list.

m0uss3:
It seems we are not on the same page about the field name.

What about using "Last updated" for remote date field and "Last watched/read" for local history date field, depending on media type, and adding a tooltip to the column header:

test-min

z411:
I like it. That said, could you keep the "Last Updated" column in this PR, and move the relative times + automatic timer to a different PR? That way we can also implement that feature in start/finish dates as well and make it configurable.

Issue #751, opened 2025-01-10
[Feature Request] Support detecting anime under a directory


barraIhsan:
While the usual format of downloaded anime usually includes the anime name (either romaji or english title) in the filename itself, so e.g. `Frieren - S1E01.mkv`, there's however some *provider* that distribute anime with the anime name only in the directory and not in the filename, for example
```
Wistoria - Wand and Sword S01 1080p Dual Audio WEBRip DD+ x265-EMBER
  - S01E01-Like a Lone Sword [E2514310].mkv
  - S01E02-As Though Undaunted [5465F323].mkv
  - S01E03-Order & Watcher [E74E7AEE].mkv
  ...
```
In such cases, the filenames contain only the season and episode information, while the anime name is present in the directory name. It would be very helpful to add support for detecting anime names from directory names.

The other benefit of detecting anime name in the directory name itself is also beneficial if the filename uses a weird formatting, for example
```
0Otakudesu_ToniKawa--01_720p.mp4
```
We can easily put all of those files under a single directory named
```
TONIKAWA Over The Moon For You
```
That way, trackma could find the files easier by searching the directory names first

barraIhsan:
![image](https://github.com/user-attachments/assets/cef1384f-d889-4426-acb1-280c120eb7c5)
Owh, I'm dumb.

Support for title metadata would be really helpful tho, im changing the title

FichteFoll:
Which tracker are you currently using and which OS? The mpris tracker prefers the title currently and I think plex and jellyfin do as well. For the filename-based ones, suche as inotify and polling, we could probably add metadata inspection via `mediafile`, an external dependency.

Either way, I occasionally get the opposite situation where the file name is easier to recognize because the title has the episode name in it, so I've been meaning to make a configuration option for it but haven't gotten around to it because it requires touching all UIs, which I've been trying to avoid unless necessary (I personally just use the CLI).

barraIhsan:
> Which tracker are you currently using and which OS? The mpris tracker prefers the title currently and I think plex and jellyfin do as well. For the strictly file-based ones, suche as inotify and polling, we could probably add metadata inspection via `mediafile`, an external dependency.
> 
> Either way, I occasionally get the opposite situation where the file name is easier to recognize because the title has the episode name in it, so I've been meaning to make a configuration option for it but haven't gotten around to it because it requires touching all UIs, which I've been trying to avoid unless necessary (I personally just use the CLI).

[...]

Issue #552, opened 2021-03-12
Jellyfin tracker only picks episode name


MyFilippo94:
I've succeffully setup integration with Jellyfin, but when the player completes the episode, Trackma asks me to add a new serie with the episode's name. In other words, Trackma receives from Jellyfin only the episode name, and isn't able to recognize the serie's name, season and episode number.

Jellyfin is organized according to The TVDB standard, as an example:
 - Monogatari
   - Season 2
     - Monogatari - S02E01 - Karen Bee - Part 1.mkv
     
As a result, it of course is unable to automatically update the list's progress, as it can't find a "Karen Bee - Part 1" serie.

I did look around in the settings, but nothing suggests I might've missed a configuration. Can I solve it by myself, or is it a software limitation?

FichteFoll:
Should be addressed by #547. Can you test that locally?

FichteFoll:
Note that you'll have to add an altname for "Monogatari 2" since there's no way trackma could deduce the correct anime title from that naming.

MyFilippo94:
I am afraid it didn't work. I admit I am not used to installing pr from github, so what I did was download the main branch, and then the three modified files from #547, merge them on the folder (respecting the paths of course), and then proceed with installation instructions (removing the previous installation first).

Launching it from the terminal still shows that it didn't manage to fetch the serie's name. I moved it to my current anilist progress to test if it'd increase by 1, so the test file was "Monogatari/Season 3/Monogatari - S03E13 - Nadeko Medusa - Part 1 Bluray-1080p.mkv". I did add "Monogatari 3" as alternate name to "Monogatari Series: Second Season" in Trackma, although it wasn't useful, as there's no trace of the season's name in the readout. The end result was the same, as by settings the program opened a dialog asking me if I wanted to add "Nadeko Medusa - Part" serie to my list.

This is the output:
```
Engine: Trackma v0.8.4 - using account <username>(anilist).
Engine: Reading config files...
Data: Initializing...
libanilist: Initializing...
Data: Using libanilist (anime)
libanilist: Downloading list...
Engine: Parsing redirection file...
Engine: Scanning local library...
Tracker (Jellyfin): Initializing...
Tracker (Jellyfin): Using Jellyfin.
Ready.
Tracker (Jellyfin): Will add Nadeko Medusa - Part 1
Tracker (Jellyfin): Waiting for the player to close.
```

I'm using Monogatari as an example, but I think none of other shows updated succeffully.
[...]

PR #814, opened 2026-01-15
Add "Prefer media title" setting (mpris only, for now)

FichteFoll:
The current mpris implementation prefers the media title as reported by the respective dbus provider (i.e. the player), but personally I found the file name to be much more accurate when matching against my library, which is why I've basically had this feature disabled locally ever since I rewrote the mpris tracker. As a result and from this experience, the setting is disabled by default.

I always wanted to upstream this but not without also making it configurable, so here goes.

This feature is not complete, however, because:

1. I don't understand the Gtk UI (never worked with it), so I didn't add it there. I assume I'd need to use a GUI editor to get the layouting right.
1. For file-based trackers, we would probably want to add something that can read media info from the files. I was planning to use `mediainfo --Output=JSON` for this, but it seems questionable whether this is available on the PATH for Windows users. Do we need to make the path configurable?
1. What about other trackers? Do Jellyfin or Plex even support retrieving the media's file name? If supported, does the default of preferring the file name make sense for them?

Edit: Jellyfin seems to support it (via https://github.com/z411/trackma/issues/552#issuecomment-799649524). 

---

I also implemented the `library_full_path` setting because that was also missing.

Related: #751
Related: #552

### DIFF
--- a/trackma/tracker/mpris.py
+++ b/trackma/tracker/mpris.py
@@ -135,6 +135,7 @@ async def collect_names(router) -> Dict[str, str]:
 
 @dataclass
 class Player:
+    config: dict[str, Any]
     router: DBusRouter
     wellknown_name: str
     unique_name: str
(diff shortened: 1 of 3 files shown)


z411:
I would also personally lean into removing the option altogether, especially if the experience shows that it works better with the filename every time.

Although it's really interesting that Jellyfin provides the show name already - maybe the Jellyfin tracker could skip the filename tokenizer altogether.

PR #832, opened 2026-02-28
mpris: Prefer filename & implement `library_full_path` setting

FichteFoll:
The full path setting causes almost none of my files in the library to be matched but I confirmed that the mpris implementation behaves identical to the library scan result at least.

Supersedes #814.
Related to https://github.com/z411/trackma/issues/751.
Related to https://github.com/z411/trackma/issues/552.


z411:
LGTM, thank you.

PR #849, opened 2026-05-31
Fix episode detection for certain batch formats when "take subdirectory name into account" is enabled

### LINKED ISSUE #848: Episode number extracted from directory name instead of filenames
# Current problem
When the "Take Subdirectory name into account" option is checked, some directory names used in batches will interfere with the episode numbering.
 `[HorribleSubs] Oshiete! Galko-chan (01-12) [1080p] (Batch)/[HorribleSubs] Oshiete! Galko-chan - 08 [1080p].mkv` will correctly detect the show name but not attribute an episode number.

# Suggested fix

Currently the full path is used for both name search and episode  search.
```
anime_info = self.parser_class(self.msg, filename)
show_title = anime_info.getName()
(show_ep_start, show_ep_end) = anime_info.getEpisodeNumbers(True)
```
My suggestion is to use the full path for name search, and switch to filename for episode search. In my case the problem is solved with this change, but it may cause new problems for other releases.
```
anime_info = self.parser_class(self.msg, filename)
show_title = anime_info.getName()
ep_info = self.parser_class(self.msg, os.path.basename(filename))
(show_ep_start, show_ep_end) = ep_info.getEpisodeNumbers(True)
```

Apcty:
Fixes #848

When library_full_path is on, _add_show_to_library extracted both series name and episode number from a file's full path, where the directory's episode range could interfere: a library with multiple batch folders ("Series (01-12)[batch]/Series - 07.mkv" and "OtherSeries [batch]/05 - Episode Title.mkv") would require this option, which causes wrong episode detection as the parser picked up the range from the directory name instead of the episode number from the filename.

Fix this by splitting the two concerns: run the parser on the full path to get the show title (benefiting from subdirectory context), but run a second parse on the bare basename to extract the episode number. Only fall back to the full-path parse for the episode number if the basename yields nothing (for a batch like "Series[batch]/07/Episode.mkv".

Note: extracting the basename via os.path.basename is slightly awkward since fullpath is available in add_to_library which calls _add_show_to_library, but this avoids larger changes to the codebase.

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -868,7 +868,16 @@ class Engine:
             # If it fails, cache it as None.
             anime_info = self.parser_class(self.msg, filename)
             show_title = anime_info.getName()
-            (show_ep_start, show_ep_end) = anime_info.getEpisodeNumbers(True)
+            if self.config['library_full_path']:
+                ep_info = self.parser_class(self.msg, os.path.basename(filename))
+                (show_ep_start, show_ep_end) = ep_info.getEpisodeNumbers(False)
+                if show_ep_start is None:
+                    (show_ep_start, show_ep_end) = anime_info.getEpisodeNumbers(True)
+                else:
+                    show_ep_start = int(show_ep_start)
+                    show_ep_end = int(show_ep_end) if show_ep_end is not None else show_ep_start
+            else:
+                (show_ep_start, show_ep_end) = anime_info.getEpisodeNumbers(True)
             if show_title:
                 show = guess_show(show_title)
                 if show:


FichteFoll:
The better fix would probably be to change the parser implementation and make it always accept a "full" path so that the parser can decide (based on its capabilities & the user settings) on how to handle the given path, which would not only make this workaround work for library scanning but also for tracker detection, but that would obviously be much more work and require some refactoring. (Ultimately, the parser instance should also be shared among multiple invocations but the architecture just doesn't work like that currently.)

Seems fine as a quick band-aid fix for me, if this solves your pain, but I'll leave it to z411 to decide whether this should be merged as-is.

PR #852, opened 2026-07-07
parser: add some hardening for anitopy & misc cleanups

FichteFoll:
~~The primary motivator here was fixing a couple exceptions with the anitopy parser I collected over the years. Since the code is already terrible (for my standards), I really didn't want to spend much time with it and let some LLM go ham instead.  The added tests help with verifying that the code works "as expected", where "as expected" is not necessarily the best solution because of issues with anitopy but should at least cover some regressions.~~

I included some other changes I did earlier since I don't think they are no-brainers and do not deserve their own merge request.

Does not fix 843 anymore.
Related: #844

### DIFF
--- a/trackma/engine.py
+++ b/trackma/engine.py
@@ -1067,7 +1067,7 @@ class Engine:
         try:
             return str(path.relative_to(searchdir))
         except ValueError:
-            return path.basename
+            return path.name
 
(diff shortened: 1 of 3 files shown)


FichteFoll:
Dropped commits that were created with code from LLMs at the request of @z411.

PR #853, opened 2026-07-07
cli: add account edit mode

FichteFoll:
Allows to re-generate OAuth codes when they expired, like Anilist's which are only valid for a year and I'm already on my 6th "account" because of this.

### DIFF
--- a/trackma/accounts.py
+++ b/trackma/accounts.py
@@ -61,7 +61,7 @@ class AccountManager:
         self.accounts['next'] += 1
         self._save()
 
-    def edit_account(self, num, username, password, api, extra={}):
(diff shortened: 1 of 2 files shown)


z411:
Looks good to me, although I'm not sure if AniList has a way to refresh its token? Does it expire after it being unused?

FichteFoll:
It just invalidates the token after 1 year and you need to generate a new one, which is what this code does since it uses the same mechanism as the add account workflow.

Although, admittedly I didn't test this, but the code change looks simple enough.

FichteFoll:
I just tested it and confirmed it works as expected. Merging.

PR #854, opened 2026-07-08
polling: fix lsof arguments

### LINKED ISSUE #851: tracker misidentified qbit as a media player, attempts to update the list
Whenever a torrent is active, tracker (polling) assumes the file is being played and attempts to update the list. Issue drops instantly when the torrent is put on pause. 

I'm using qt version 0.10.3 built with poetry.

Image

FichteFoll:
Different selection filters are OR-combined by default, which means the command was matching for files in the provided path OR files opened by commands that match the pattern. The `-a` option is what causes the filter to be AND-combined.

Additionally, suppress warnings (`+w`) and repeat the `+D` option for all watch dirs (multiple filters for the same kind are OR-combined even with `-a`) to implement the TODO comment and use a single command.

Fixes #851

### DIFF
--- a/trackma/tracker/polling.py
+++ b/trackma/tracker/polling.py
@@ -26,21 +26,27 @@ class PollingTracker(tracker.TrackerBase):
     name = 'Tracker (polling)'
 
     def get_playing_file(self, watch_dirs, players):
+        cmd = [
+            'lsof',
+            '-a',  # cause selection options to be AND-combined
+            '-w',  # suppress warnings (the man-page is ambiguous here but it's -w)
+            '-F', 'n',  # output only the name
+            '-c', f'/{players}/',  # match process (extended regex)
+        ]
         for path in watch_dirs:
-            # TODO: We'll run lsof once for each directory for now.
-            try:
-                lsof = subprocess.Popen(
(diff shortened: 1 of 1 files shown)


FichteFoll:
I'm really not sure about the warnings flag, tbh. The man page is a bit ambiguous here, talking about enabling the *suppression* of warnings initially but the later paragraph does not mention suppression. 69d4d2bb1936fdaf651f312ab35187e01d32d20f specifically changed the flag from +w to -w.

FichteFoll:
Okay, I confirmed the behavior and `-w` is correct.
I verified this with:

- `lsof -b -w 2>&1 | grep WARNING`
- `lsof -b +w 2>&1 | grep WARNING`

and the `+w` line showed a lot of warnings while `-w` showed none.

Here is the man text verbatim, by the way:

```
     +|-w     Enables (+) or disables (-) the suppression of warning messages.

              The lsof builder may choose to have warning messages disabled or enabled by default.  The default warning message state is indicated in the out‐
              put of the -h or -?  option.  Disabling warning messages when they are already disabled or enabling them when already enabled is acceptable.

              The -t option implies the -w option.
```

PR #855, opened 2026-07-11
Refactor MPRIS state tracking and improve tracker logic

FichteFoll:
What started out as a simple idea for decoupling mpris dbus and the tracker state & logic turned out to become a slightly larger refactoring on the whole tracker architecture that ended up fixing two not really important but long-standing issues I had noticed with the mpris tracker over the past months where it would be spamming a lot of state changes in the debug log when the *relevant* state really didn't change much.

In the end, I spent quite a bit of time for relatively little gain, but I would still say that the code is now in a much better state to reason about and the added type annotations will also help in the long run.

### DIFF
--- a/trackma/tracker/inotify.py
+++ b/trackma/tracker/inotify.py
@@ -73,7 +73,7 @@ class inotifyTracker(inotifyBase.inotifyBase):
                     # Default blocking duration is 1 second
                     # This will count down like inotifyx impl. did
                     self.update_show_if_needed(
-                        self.last_state, self.last_show_tuple)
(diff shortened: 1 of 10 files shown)


FichteFoll:
Hm, this needs some refinement. The timer is not properly reset when a new episode is found. Will fix later.

Edit: fixed in https://github.com/FichteFoll/trackma/commit/b94c4f9c323d9f7c5415ac9d4c47e698cd25a4ae

FichteFoll:
Redacting this PR because it contains a lot of code generated by LLMs which @z411 has expressed rejection for.

I will continue this work in my fork.

Issue #826, opened 2026-01-21
python-pyqt6 and urllib3 dependencies in Fedora


dyskette:
Hi, there.

When packaging the newest release I noticed two things:

1. `urllib3` requires `^2.6.0` version but we have `2.3.0` in [Fedora 42](https://packages.fedoraproject.org/pkgs/python-urllib3/python3-urllib3/fedora-42.html]. I didn't find any usage of it in the codebase, so is it possible to remove the dependency?
2. `pyqt6` requires `^6.10.2` version, which is not available in any of the current supported Fedora releases. We have:
    - `6.9.1` in Fedora 42
    - `6.10.1` in Fedora 43
    - `6.10.1` in Fedora rawhide
    - [python3-pyqt6 package](https://packages.fedoraproject.org/pkgs/python-pyqt6/python3-pyqt6/)

I don't know much about qt6, so my question would be if we can relax the version requirement.

dyskette:
If not possible, then I think I could wait for the new pyqt6 version to drop in Fedora 43 and Rawhide. Then release the new version of trackma for those only. And not updating trackma in Fedora 42.

dyskette:
Awesome. Then it seems we are good. I can patch the release for now in the package itself to make the latest release available in Fedora.

PR #857, opened 2026-07-12
Switch to uv for version management

FichteFoll:
As discussed in #826 already, we wanted to switch from poetry to uv, so here it is.


FichteFoll:
Redacting this PR because it contains a lot of "code" generated by LLMs which @z411 has expressed rejection for.

